PR #24, opened 2017-07-20
Select all fields if parent type is an interface

### LINKED ISSUE #20: Problems with interfaces
I have a problem accessing fields in types implementing an interface. 
I have a simple interface 'TestInterface' defining fields type and field1. I have an object TestObject implementing the interface and defining the fields type, field1 and field2.

I have a simple query returning a list of TestInterface type objects.

The schema generates correctly and looks fine (inspecting in graphiql).

I do a simple query to confirm that types are being resolved correctly:

{
  Tests{
    __typename
    field1
  }
}

which is confirmed in result:

{
  "data": {
    "Tests": [
      {
        "__typename": "TestObject",
        "field1": "asd"
      },...
    ]
  }
}

However if I try to query field2, which is defined on TestObject (but not in the interface):

{
  Tests{
    __typename
    field1
    ... on TestObject {field2}
  }
}

I get the following: Field "field2" is not defined for type "TestInterface\"

So even though the first query confirms that the return is of concrete type TestObject, for some reason it tries to resolve field2 on TestInterface.

RSchwan:
Fixes #20

### DIFF
--- a/src/Rebing/GraphQL/Support/SelectFields.php
+++ b/src/Rebing/GraphQL/Support/SelectFields.php
@@ -3,6 +3,7 @@
 namespace Rebing\GraphQL\Support;
 
 use Closure;
+use GraphQL\Error\InvariantViolation;
 use GraphQL\Type\Definition\InterfaceType;
 use GraphQL\Type\Definition\ListOfType;
 use GraphQL\Type\Definition\ResolveInfo;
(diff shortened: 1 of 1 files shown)


RSchwan:
ping @rebing

src/Rebing/GraphQL/Support/SelectFields.php
```diff
@@ -115,7 +116,12 @@ protected static function handleFields(array $requestedFields, $parentType, arra
                 continue;
             }
 
-            $fieldObject = $parentType->getField($key);
+            // If field doesn't exist on definition we don't select it
+            try {
+                $fieldObject = $parentType->getField($key);
+            } catch (\Exception $e) {
```
rebing: I don't think this the correct solution. Maybe check, if the field exists or catch the specific exception?

PR #27, opened 2017-08-06
Add artisan make commands

RSchwan:
Mostly copied from https://github.com/Folkloreatelier/laravel-graphql


src/Rebing/Console/stubs/mutation.stub
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    public function args()
+    {
+        return [
+
+        ];
+    }
+
+    public function resolve($root, $args, $context, ResolveInfo $info)
```
rebing: SelectFields should be here as well

PR #37, opened 2017-08-22
Add enum support

jbbr:
This is just a copy from https://github.com/Folkloreatelier/laravel-graphql/blob/master/src/Folklore/GraphQL/Support/Type.php but it works fine.

So it's documentation applies. See https://github.com/Folkloreatelier/laravel-graphql/blob/58ca76dd1cf50e099e3f3d317e699078b400a289/docs/advanced.md#enums

I could copy the docs from there or write it from scratch if you prefer.

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -8,13 +8,14 @@
 - [Type relationship query](#type-relationship-query)
 - [Pagination](#pagination)
 - [Batching](#batching)
+- [Enums](#enums)
 
 ### Authorization
 
 For authorization similar to Laravel's Request (or middleware) functionality, we can override the `authorize()` function in a Query or Mutation.
 An example of Laravel's `'auth'` middleware:
 
-```
+```php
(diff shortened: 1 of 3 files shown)


jbbr:
I added some documentation.
Slightly modified version of https://github.com/Folkloreatelier/laravel-graphql/blob/58ca76dd1cf50e099e3f3d317e699078b400a289/docs/advanced.md#enums

docs/advanced.md
```diff
@@ -505,9 +505,10 @@ First create an Enum as an extension of the GraphQLType class:
 // app/GraphQL/Enums/EpisodeEnum.php
 namespace App\GraphQL\Enums;
 
-use RebingFolklore\GraphQL\Support\Type as GraphQLType;
```
jbbr: whoops sorry about that

jbbr:
thanks

Issue #31, opened 2017-08-11
Help! How to rename field in query?


MrJmpl3:
I need rename field in query and in backend rename to field 

Same Laravel Fractal

Example:
My table users: Columns: Id, Name , Email

Request:
query Users { users{ code nick contact} }

I dont need exposed the field from my database ..

Any ideas?

rebing:
Use a magic attribute `getCodeAttribute()` in the UserModel (or `resolve` in GraphQL Type) and then `selectable => false` in the Type. https://github.com/rebing/graphql-laravel#creating-a-query

nyalex:
@rebing I am also stuck on this. Could you elaborate on how to do this? Your solutions seem to be hacks, unfortunately. Perhaps it's a good idea to add a "name" key in the `fields()` method in the type file?

rebing:
There _is_ a 'name' attribute in fields() type

mebranemlm:
Hi. Is there a way to use `$field->getSelect()` together with aliases? I mean when I use "name"  attribute in fields() type it returns the value with the alias attribute (for example `'nick'=> 'Jon Doe'` instead of `'name'=> 'Jon Doe'`, using MrJmpl3 example) and that's ok, but when I want to use $field->getSelects() in Query and 'selectable'=>true in Type, it returns `'users.nick'` instead of `'users.name'`.  And if I use 'selectable' => false it simply doesn't return the field.

Same thing with $args in Query and Mutation. When I use name attribute in args() function (for example: `'email' => ['name'=>'contact']`), resolve() function receives `[contact=>"jondoe@gmail.com"] `instead of `[email=>"jondoe@gmail.com"]`. And if I use 'contact'=>['name'=>'email'] the interface requires email instead of contact. The idea is that when I send alias attributes from interface, in resolve() I want to get the original attributes. 

I'm using foreach to iterate in Type fields to transform $args and select aliases to original attributes, but it complicates too much when I want to create iterate functions for where, sort, etc. Maybe is there a better way to do this?

rebing:
Hi, I'm not really familiar with the `name` attribute and how it _should_ work. You can maybe look into it and create a PR, if you find the problem.

PS. `'selectable' => false` is working correctly and `getSelect()` should NOT return the field in that case.

PR #38, opened 2017-08-24
Add parameters support to instance custom fields

emiliogrv:
* added custom field for dates
* added support for the use of parameters in the constructor of any custom-field

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -369,7 +369,7 @@ When creating a mutation, you can add a method to define the validation rules th
 		{
 			return [
 				'id' => ['name' => 'id', 'type' => Type::string()],
-				'email' => ['name' => 'password', 'type' => Type::string()]
+				'email' => ['name' => 'email', 'type' => Type::string()]
 			];
 		}
 		
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -161,9 +161,9 @@ use Rebing\GraphQL\Support\Field;
 
 class PictureField extends Field {
         
-        protected $attributes = [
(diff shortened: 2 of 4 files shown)


emiliogrv:
Hello,
Is there a problem with this?
I also have another PR to solve this: https://github.com/rebing/graphql-laravel/issues/31 as an alias

rebing:
Sorry, but I don't like a separate field for DateTime, because it can vary so much and I don't think it's necessary

emiliogrv:
Hello,
The DateTime was really an example, I can delete it and make another commit, the idea of PR is to give the opportunity to send parameters to the builder of any custom-field if necessary

rebing:
I still don't quite understand what the PR is trying to accomplish here. What is purpose of the `field` attribute?

emiliogrv:
The attribute `field` is for send the class of custom-field and attribute `params` is for send parameters to constructor funtion of that custom-field
This PR is for that, send parameters to constructor of any custom-field  if need. The DateTime custom.field is only as example of that feature

rebing:
Why not just use `'created_at'` => DateField::class`? What is the use of `params`? Has another package done something like this?

emiliogrv:
> Why not just use 'created_at' => DateField::class?
Because can't send values to constructor of class

> What is the use ofparams`? 
Config any options in the class of custom-field for the internal logic, the name of field in the database, for example for dates, many names of dates can be resolve with only one custom-field class

> Has another package done something like this?
No, that's why I propose this solution, I had the same need in another work project, and that's why I share it

PR #77, opened 2018-01-25
Update GraphQLServiceProvider.php

codeitlikemiley:
Only Load Make Commands In Local and Staging Environment


rebing:
This is way too specific

PR #78, opened 2018-01-25
remove laravel

codeitlikemiley:
remove laravel dependencies so lumen can use it


rebing:
Laravel is a required dependency, we cannot remove it

PR #81, opened 2018-02-08
Added 5.6 Laravel dependency option

rebing:


fgreinus:
Could you please also tag a new version? Would be awesome in order to properly update everything

PR #83, opened 2018-02-24
Fixing undefined method with embedded array of objects

JosephMaxwell:
*Example Query*
```
query {
  Me {
    email, first_name, last_name, agreed, subscriptions {
      name
      start
      end
      trial
    }
  }
}
```

*Type Definition*
```
...
'subscriptions' => [
    'type' => Type::listOf(\GraphQL::type('Subscription')),
    'description' => 'Active subscriptions'
],
...
```

I found that the `$parentType` does not have a `getFields` method on it. By just bypassing this call for a `ListOfType` class, the problem is solved.


JesusMayuri7:
please, example completed

PR #84, opened 2018-02-26
Allow subscription types to be added

BasMulders:
Based on @petecoop's PR at the @Folkloreatelier repo (https://github.com/Folkloreatelier/laravel-graphql/commit/70707eca85eec1a41a7cb9061053feee768a69d6).

### DIFF
--- a/src/Rebing/GraphQL/GraphQL.php
+++ b/src/Rebing/GraphQL/GraphQL.php
@@ -46,6 +46,7 @@ class GraphQL {
 
         $schemaQuery = array_get($schema, 'query', []);
         $schemaMutation = array_get($schema, 'mutation', []);
+        $schemaSubscription = array_get($schema, 'subscription', []);
         $schemaTypes = array_get($schema, 'types', []);
 
         //Get the types either from the schema, or the global types.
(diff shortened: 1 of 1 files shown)


petecoop:
HI

rebing:
@BasMulders Could you please write a short description about what subscriptions are? I couldn't find anything in Folkloreatelier's repo.

BasMulders:
@rebing Sure. Subscriptions in GraphQL enable the server to push data to the clients. If you're working with real-time or nearly real-time data, you need to update the data at the client side. An option here would be to make use of polling, but a more efficient way is to use subscriptions. In Laravel, we can use the event/broadcasting system to send certain updates to the client.

Ideally, one should use websockets to keep an 'open' connection with the client. Personally, I do not think PHP is suitable for doing this. An option here is to set up a nodejs server that does this. However, this would require us to define our GraphQL scheme two times (in our Laravel GraphQL and our nodejs server). We can solve this by importing the GraphQL scheme for the nodejs server from Laravel (introspection). But that requires us to have a scheme that supports subscriptions. Hence the need for subscriptions in this package.

[Graphql-php](http://webonyx.github.io/graphql-php/type-system/schema/#configuration-options), the package on which Folkloreatelier's repo is based, supports this as well: "Currently presented for compatibility with introspection query of graphql-js, used by various clients (like Relay or GraphiQL)".

petecoop:
[Subscriptions in GraphQL spec](http://facebook.github.io/graphql/draft/#sec-Subscription). A Subscription is just a query but with the added expectation that this will set up an "Event Stream" that will send updates if they query changes.

How the actual streaming is done is entirely up to your implementation, as @BasMulders explains above PHP probably isn't the right place to handle this, and especially with Laravel it could be done through [Broadcasting](https://laravel.com/docs/5.6/broadcasting), and use something like Pusher or a separate node.js server (e.g. [laravel echo server](https://github.com/tlaverdure/laravel-echo-server)) to handle sending the event to the client.

I think this is an area that we're missing currently with GraphQL in PHP and could really use some thought into building it so we can have functionality that's on par with node.js. Obviously allowing the Subscription type to be added is a first step

rebing:
I knew about subscriptions and thought webonyx had actually implemented it to the PHP package already. Fair enough though, if it helps build the schema and doesn't seem to break anything.

Issue #65, opened 2017-12-21
alias function


crissi:
```
    class CaseType extends GraphQLType {
        
        protected $attributes = [
            'name'          => 'Case',
            'description'   => 'A Case',
            'model'         => Order::class,
        ];

        public function fields()
        {
            return [
                'id' => [
                    'type' => Type::nonNull(Type::int()),
                    'description' => 'The id of the user',
                ],
                'case_number' => [
                    'type' => Type::nonNull(Type::string()),
                    'description' => 'An unique identifier for the case',
                    'alias' => 'number'
                ],
                'caseType' => [
                    'type'          => GraphQL::type('caseType'),
                    'description'   => 'type of case',
                    'alias' => 'type'
                ]
            ];
        }
    }
```

I would like to implement a alias function.

The output will send the alias instead of the key.

The input will receive the alias and convert it to the correct key value, so the queries will work correctly

Would you be interested in merging a pullrequest with that functionality?

rebing:
What about using the `'name'` attribute instead?

crissi:
![image](https://user-images.githubusercontent.com/3076578/34255565-99e1c28c-e651-11e7-81be-bb0d1d86347b.png)

I was looking at what your library is build on and it was using name and key interchangeable: http://webonyx.github.io/graphql-php/type-system/input-types/

But makes sense to use that instead, since we already have the the key for the internal name.

PR #85, opened 2018-03-08
Use a a different column name when selecting a field

romainprevost:
Hello!

I have just started using your package but I'm using a legacy database with non consistent columns naming conventions.

This simple PR adds the possibility to set the column name of a field when it is linked to an Eloquent model, so that the fields name can be different than the columns name in the Eloquent table.

I've made the attribute key configurable and I also updated the UserType example.

### DIFF
--- a/src/Rebing/GraphQL/Support/SelectFields.php
+++ b/src/Rebing/GraphQL/Support/SelectFields.php
@@ -198,6 +198,10 @@ class SelectFields {
                 // Select
                 else
                 {
+                    // If an alias for the field is set, use it as DB column
+                    $columnNameProperty = config('graphql.column_name_attribute');
+                    $key = isset($fieldObject->config[$columnNameProperty]) ? $fieldObject->config[$columnNameProperty] : $key;
+
                     self::addFieldToSelect($key, $select, $parentTable, false);
 
                     self::addAlwaysFields($fieldObject, $select, $parentTable);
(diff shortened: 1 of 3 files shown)


patrickreck:
We had to implement this in our system too when we started using GraphQL.

+1 for this

rebing:
#65 I didn't see the point in defining a variable name for the alias key

romainprevost:
It is not quite the same as issue #65, if you use the name property for a field, the name will indeed be used in SelectFields but it will also be used in the response.

My practical use case is this: 
I have a user table with an **id** and **login** column for the username. 

I want want my User type to have an **id** and **name** field to be more consistent. 
With the default type description, in the resolve function, SelectFields will return "name" as column for the User which does not exist in the table

```php
    public function fields()
    {
        return [
            'id'    => [
                'type'          => Type::nonNull(Type::id()),
                'description'   => 'Id of the player'
            ],
            'name'    => [
                'type'          => Type::nonNull(Type::string()),
                'description'   => 'Name of the player'
            ],
        ];
    }
```

If I add `name => 'login'` to the name field, the correct column will be selected in the resolve function, but the schema is modified as well and login will be returned instead of name and you can't query for name but for login instead.

I understand that it is not needed for everyone but if you work with a legacy database or if you don't want to have the table column names in your schema it can be useful and you keep the very nice SelectFields feature.

You can achieve the same thing in the resolve function with a check for queried field and change the query accordingly but it can be cumbersome if you have to do it for many fields or many types.

PR #86, opened 2018-03-10
Alias field

rebing:


jbbr:
I really like the idea of the alias field but the implementation feels incomplete to me as the field is still resolved by its graphql type name which requires a custom `resolve` or `getXYAttribute` method.

I would expect the example:
```php
                'id' => [
                    'alias' => 'user_id',
                ],
```

to resolve `id` by using `$root->user_id`.
Instead the default resolver still tries to access `$root->id` which results in the field to be `null`.
So I have to add:

```php
                'id' => [
                    'alias' => 'user_id',
                    'resolve' => function ($root) { return $root->user_id; }
                ],
```

which seems verbose.

Is this behavior intended?

Issue #63, opened 2017-11-11
Union type


DiederikvandenB:
Another question. I was trying to build a search query that goes through multiple models. I was looking at this GraphQL Type: https://github.com/webonyx/graphql-php/blob/master/docs/type-system/unions.md

Is it possible to implement this using this package?

Basically, this is what I want to achieve:
https://medium.com/the-graphqlhub/graphql-tour-interfaces-and-unions-7dd5be35de0d

Cheers!

rebing:
I haven't worked with UnionTypes, but it _should_ already work, since this package is built on top of the webonyx package that you linked. Have you tried creating a UnionType and adding it as a field?

DiederikvandenB:
Thank you for your response. I have tried the following:
https://gist.github.com/DiederikvandenB/1beee6e88d3671e72f2c639f17d10664

Note that I only register the `SearchQuery` in the `config/graphql.php`.

Attached to the gist is also the error message. What am I doing wrong?

Cheers!

DiederikvandenB:
Months later I am revisiting this issue. Any points you could give me?

jbbr:
@DiederikvandenB you can try my PR #91 
I added basic support for Unions to the Rebing Type class. It does work despite the WIP flag.

jbbr:
Another hint @DiederikvandenB : I think you can't add raw webonyx types to the Facade/graphql config.
But you can indeed reference instances of your custom type in other types and queries or even defined inline.
e.g:

```php
'fields' => [
  'fieldOther' => [ 'type' => GraphQL::type('SomeRebingType')],
  'fieldUnion' => ['type' => new UnionType([ 'types' => [...]],
]
```

jbbr:
@DiederikvandenB You can try the latest Version release 1.12.0, the PR has been merged.
There is also some documentation on https://github.com/rebing/graphql-laravel/blob/master/docs/advanced.md#unions

PR #91, opened 2018-04-18
Add support for Unions

jbbr:
This PR improves/fixes support for Unions:

- simplifies the definition of Union Types when using the rebing `Support\Type` base class
- skip SelectFields for union types (same as for Interfaces)
- add basic example to documentation

This is not ideal. It would be great to preserve the `SelectFields` feature. Also it might be better to use a new class `Rebing\Graphql\Support\UnionType` instead of using a boolean attribute in `Support\Type`.

So this is more for reference for others having issues with booleans (#63) and open for discussion. 
@rebing It would be great if you could review this PR and give some feedback. I'm willing to fix and improve this PR until it's ready to get merged.
Thanks

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -83,6 +83,7 @@ config/graphql.php
 - [Pagination](docs/advanced.md#pagination)
 - [Batching](docs/advanced.md#batching)
 - [Enums](docs/advanced.md#enums)
+- [Unions](docs/advanced.md#unions)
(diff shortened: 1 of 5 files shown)


jbbr:
I moved the union-changes to an extra class `UnionType` and added some documentation.
Please check again @rebing, thanks!

PR #92, opened 2018-04-18
Add Support for FieldDefinition instance in fields array

jbbr:
This PR allows using instances of `GraphQL\Type\Definition\FieldDefinition` in the `fields` array of a type.
This is very useful when using Interfaces as the `InterfaceType` has a `getField` and `getFields` method for sharing interface fields.

This is documented on http://webonyx.github.io/graphql-php/type-system/interfaces/#sharing-interface-fields

Example:

```php
class SubType extends GraphQLType
{
    protected $attributes = ['name' => 'SubType'

    function interfaces()
    {
        return [GraphQL::type('SomeInterface')];
    }

    public function fields()
    {
        $interfaceFields = GraphQL::type('UserInterface')->getFields();
        $additionalFields = [
            'special_field' => Type::string(),
        ];
        array_merge($interfaceFields, $additionalFields);
    }
```
or:
```php
    public function fields()
    {
        $interface = GraphQL::type('UserInterface');
        return [
            $interface->getField('id'),
            $interface->getField('name'),
            'special_field' => Type::string(),
        ];
    }
```

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -83,6 +83,7 @@ config/graphql.php
 - [Pagination](docs/advanced.md#pagination)
 - [Batching](docs/advanced.md#batching)
 - [Enums](docs/advanced.md#enums)
+- [Interfaces](docs/advanced.md#interfaces)
 
 ### Schemas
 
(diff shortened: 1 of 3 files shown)


jbbr:
I added some documentation for Interfaces (Mostly copied from https://github.com/Folkloreatelier/laravel-graphql/blob/develop/docs/advanced.md#interfaces ) and added some snippets for shared interface fields usings `getField`

Issue #26, opened 2017-08-02
Input Object Type (Validation Rules) & Input Object Type name collision


swinyx:
#FIRST ISSUE

When am creating a mutation that receives the following (where id a the top level is required, while id at the nested object is sometimes required, the top level validation fails unless I renamed the inner object 'id' key to something else.

Also, how to set validation rules for input object type args ?

For example, I am trying to mutate an object with the following 3 properties (1 property is an input object type)
id - required (Int)
name -sometimes required (String)
subjects - sometimes required(InputObjectType), but if present{
 - id (sometimes required)
- name (required)

}

`
{
  "id": 1,
  "name": "Samle Content 1",
  "subjects": [
    {
      "id": : 1,
      "name": "Sample content 2",
    },
    {"name": "Sample Content 3"}
  ]
}`

My args() and rules() function in the mutation looks like the following. However I am not sure where to specify the validation rules for the InputObjectType

`
public function args(){

		return [
					'id' => [
						'name' => 'id',
						'type' => Type::int(),
					],
					'name' => [
						'name' => 'name',
						'type' => Type::string(),
					],
					'subjects' => [
						'name' => 'subjects',
						'type' => Type::listOf(new InputObjectType([
									    'name' => 'UpdateClassroomSubjectInput',
									    'fields' => [
									        'id' => [
									            'type' => Type::int(),
									            'description' => 'The id of the subject'
									        ],
									        'name' => [
									            'type' => Type::string(),
									            'description' => 'The name of the subject'
									        ]
									    ]
									]))
					]

					//'subjects' => Type::listOf( new \App\GraphQL\InputTypes\ClassroomSubjectInputType)

				];
	}`

`	
public function rules(){

		return [
					'id' => [
						'required'
					],
					'name' => [
						'sometimes',
						'required'
					],
					'subjects' => [
						'sometimes',
						'required'
//How to specify the nested object validation rules here ? Or should I simply validate them in the resolver ?

					]
				];
	}
`

#SECOND ISSUE

It doesn't seem there is a way to create a class name for input object types and reuse them in different mutations ( just like the field classes), The config file has no options to provide a list of input object types that can be loaded in the schema ( like the types or mutation list)

rebing:
1. I think it can currently only be done in the resolver. I think it can be solved by passing the $args into the rules($args). Feel free to make a PR, if you create a solution.

2. I've solved it by creating a new class that extends the InputObjectType and passing a unique name every time:

```
[...]

PR #94, opened 2018-04-24
Support for security rules and validation for input object type

illambo:
Reference discuss #26, the implementation is based on folkloreatelier repository:
- Add support for security rules
- Add support for validate with rules an input object type

I can not currently produce tests.


rebing:
I'm really sorry about the late reply. Could you please elaborate a little more about what the PR does?

Issue #96, opened 2018-05-14
Authorization to list a query / mutation in schema


alanwillms:
Currently we can implement an `authorize` method in order to check if the authenticated user can _execute _a query or mutation, but even if the user is not able to run it, it will be listed in the schema documentation.

It would be nice if I could specify if a query and/or mutation can appear in the schema documentation.

alanwillms:
This is our problem: we want to have only one end-point for GraphQL: `/graphql`:

* if the user is not authenticated, we will only show the `createToken` mutation
* if the user is an administrator, we will list all queries and mutations
* if the user is an analyst, we will list all queries
* if the user is a seller, we will only list mutations/queries related to sales
* etc.

rebing:
Seems like dynamic schema generation is a controversial issue. There are valid thoughts from both sides here: https://github.com/graphql/graphql-js/issues/113

I think a way for implementing it would be to add a boolean `limit_introspection_by_authorize` option to config that would remove unauthorized Fields from schema, when enabled. Not sure how easy it is to implement without performance issues, but you can give it a try.

alanwillms:
I've opened a PR (#101) so we can discuss it. The implementation suggestion I submitted is working and based on the way `graphql-ruby` works: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/warden.rb

There are no working unit tests in this repository, so I tested it against an application I work with. After we finish discussing the implementation details and we are happy with the code, I will write unit tests for it.

PR #101, opened 2018-05-16
Queries and mutations custom visibility in schema introspection

alanwillms:
See #96

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -109,7 +109,7 @@ in addition to the global middleware. For example:
             'profile' => 'App\GraphQL\Query\ProfileQuery'
         ],
         'mutation' => [
-        
+
         ],
         'middleware' => ['auth'],
     ],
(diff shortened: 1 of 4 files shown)


alanwillms:
This is not ready to be merged, I opened this PR to start a discussion

alanwillms:
I think a single method would be better too, but my concern is that `authorize` can be scoped according to the `$args` param. For example, one may feel encouraged to use `authorize` to check if the user calling `updatePost` is the owner of the post (`$post->user_id == Auth::id()`).

Well, anyway if someone does that, he or she will get an error while introspecting can fix it by put the fine-grained level authorization in the proper place (inside `resolve` maybe).

I will change to use `authorize` instead.

alanwillms:
>  and maybe have another boolean field in the config, which toggles visibility based on authorizations

Do you think that configuration should be per query/mutation or per schema?

rebing:
I think the configuration should apply to the whole schema

alanwillms:
@rebing please take a look now

rebing:
I'm sorry, but I'm not sure that's still the right course, since an empty array is passed to `authorize` instead of query arguments. Also, all authorize functions are executed, which is very costly and unnecessary in most cases.

Perhaps it should be based on middleware instead?

alanwillms:
No problem, but that's beyond my knowledge about Laravel. Can you point me to the right direction?

rebing:
I'm not exactly sure how to do it either. It's actually probably not even possible to evaluate the middleware before schema introspection.

alanwillms:
If anybody find this, there is this solution for the problem:

https://github.com/fidelize/graphql-laravel-authorize-introspection

It is not the most beautiful or clean solution, but it works.

PR #102, opened 2018-05-21
Added ability to upload files.

ddeath:
This is based on https://github.com/Ecodev/graphql-upload

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -72,6 +72,7 @@ config/graphql.php
 - [Creating a query](#creating-a-query)
 - [Creating a mutation](#creating-a-mutation)
 - [Adding validation to mutation](#adding-validation-to-mutation)
+- [File uploads](#file-uploads)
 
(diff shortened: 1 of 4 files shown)


faistoiplaisir:
Hi,

There's something I don't understand about that UploadType. Why did you copy/paste the code from Edodev/graphql-upload into your package ?

It seems to be better to have a dependance to the original package, or there's another reason ? I missed something ?

Explaining in your readme.md how to use the original package is better to copy/paste that existing module into your package I think ... no ?

ddeath:
@faistoiplaisir hi Yes, I wanted to do that. But there is a problem. You can not use it as middleware, because that middleware is modifying the request. And because it is a multipart, FormRequest in laravel, it will not reflect changes.

The second reason is that the graphql-upload package is base on PSR-7 Request interfaces so it is not directly compatible with Laravel requests...

PR #113, opened 2018-07-17
Added ability to customise validation errors

chrispage1:
Currently, with validation there is no way to customise the errors that come back. This simple update, allows a method to be created within the mutation allows the creation of a method called `validationErrorMessages`.

`validationErrorMessages` should return an array of rules. For example -

```
public function validationErrorMessages($args = array ()) {
    return [
        'role_id.required' => 'Please select a valid user role',
    ];
}
```

This is in accordance with the Laravel Validator - https://laravel.com/docs/5.6/validation#custom-error-messages

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -461,6 +461,28 @@ When you execute a mutation, it will return any validation errors that occur. Si
 	}
 ```
 
+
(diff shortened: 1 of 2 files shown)


chrispage1:
@rebing I've added documentation and updated the way `validationErrorMessages` is used (overriding now rather than defining)

PR #123, opened 2018-08-19
Support additional route group attributes

rwinzhang:
Sometimes there are multiples authentication guard defined, it is better if we configure it

### DIFF
--- a/src/config/config.php
+++ b/src/config/config.php
@@ -46,6 +46,14 @@ return [
     // Any middleware for the graphql route group
     'middleware' => [],
 
+    // Additional route group attributes
+    //
+    // Example:
+    //
+    // 'route_group_attributes' => ['guard' => 'api']
+    //
+    'route_group_attributes' => [],
+
     // The name of the default schema used when no argument is provided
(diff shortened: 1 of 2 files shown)


rwinzhang:
@rebing it is actually introduced in Laravel v5.2 –– [changelog](https://laravel.com/docs/5.2/releases#laravel-5.2).
If you want to support v5.1, we can use more general config likes, 
```
'extra_route_groups' => ['guard' => 'api']
```

rebing:
Yes, I think the `extra_route_groups` is a better solution for maintaining backwards compatibility.

rwinzhang:
@rebing I have update my commit

src/config/config.php
```diff
@@ -46,6 +46,9 @@
     // Any middleware for the graphql route group
     'middleware' => [],
 
+    // Additional route group attributes
+    'route_group_attributes' => ['guard' => 'api'],
```
rebing: This should not be added as a default option. Add it to the comment as an example instead

rwinzhang: done

PR #141, opened 2018-10-15
Format error updated

albertcito:
If you have some exception in your query or mutation this will print only the message, and a location not related. for example if I print a variable no defined, the line and column is not related with my file `TagQuery.php`

```graphql
{
  "data": {
    "tag": null
  },
  "errors": [
    {
      "message": "Undefined variable: undefinedVarTest",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ]
    }
  ]
}
```

I added more information to the error message, now it is easier to debug my code in case of error.

```graphql
{
  "data": {
    "tag": null
  },
  "errors": [
    {
      "message": "Undefined variable: undefinedVarTest",
      "code": 0,
      "line": 42,
      "file": "/project/app/GraphQL/Query/TagQuery.php",
      "class": "ErrorException",
      "trace": [
        {
          "file": "/project/app/GraphQL/Query/TagQuery.php",
          "line": 42,
          "function": "handleError",
          "class": "Illuminate\\Foundation\\Bootstrap\\HandleExceptions",
          "type": "->"
        },
   ...
}
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,13 +29,13 @@
     "type": "project",
     "require": {
         "php": ">=5.5.9",
-        "illuminate/support": "5.1.*|5.2.*|5.3.*|5.4.*|5.5.*|5.6.*",
+        "illuminate/support": "5.1.*|5.2.*|5.3.*|5.4.*|5.5.*|5.6.*|5.7.*",
         "webonyx/graphql-php": "~0.10.0",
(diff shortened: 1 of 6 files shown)


rebing:
The validation error line and column refer to the incoming GraphQL query, but you are reporting internal exceptions. You can change it by using a custom ErrorFormatter in the config file.

PR #169, opened 2019-01-10
Support custom PaginationType & Enhance Upload Type & Fix Privacy and SelectFields no working on Paginate

a70537952:
- Support custom PaginationType
- Use UploadType getInstance function, so we dont necessary to specific name for every UploadType
- Fix Privacy & SelectFields no working on Paginate

### DIFF
--- a/src/Rebing/GraphQL/Support/SelectFields.php
+++ b/src/Rebing/GraphQL/Support/SelectFields.php
@@ -44,7 +44,8 @@ class SelectFields {
             // If use pagination,
             // $parentType & $requestedFields should use QueryType
             // instead of paginationType
-            if($info->parentType->name !== 'Mutation' && $info->returnType instanceof ObjectType){
+            $paginationType = config('graphql.pagination_type', PaginationType::class);
+            if($info->returnType instanceof $paginationType){
                 // get pagination data key [default: data]
                 $paginationDataKey = array_keys($requestedFields)[0];
                 // use QueryType fields instead of pagination fields
--- a/src/Rebing/GraphQL/Support/SelectFields.php
+++ b/src/Rebing/GraphQL/Support/SelectFields.php
@@ -39,21 +39,19 @@ class SelectFields {
         {
             self::$args = $args;
             $requestedFields = $info->getFieldSelection(5);
-            $parentTypeName = get_object_vars($parentType)['config']['name'];
+            $parentTypeName = $parentType->config['name'];
 
             // If use pagination,
             // $parentType & $requestedFields should use QueryType
             // instead of paginationType
             $paginationType = config('graphql.pagination_type', PaginationType::class);
             if($info->returnType instanceof $paginationType){
-                // get pagination data key [default: data]
-                $paginationDataKey = array_keys($requestedFields)[0];
-                // use QueryType fields instead of pagination fields
-                $requestedFields = $requestedFields[$paginationDataKey];
-                // use QueryType instead of pagination type
-                $parentType = get_object_vars(
-                    get_object_vars($parentType)['config']['fields'][$paginationDataKey]['type']
-                )['ofType'];
+                foreach ($parentType->config['fields'] as $key => $field) {
+                    if($field['type'] instanceof ListOfType) {
+                        $requestedFields = $requestedFields[$key];
+                        $parentType = $field['type']->ofType;
+                    }
+                }
             }
 
             $fields = self::getSelectableFieldsAndRelations($requestedFields, $parentType);
--- a/src/Rebing/GraphQL/Support/SelectFields.php
+++ b/src/Rebing/GraphQL/Support/SelectFields.php
@@ -37,8 +37,24 @@ class SelectFields {
         if( ! is_null($info->fieldNodes[0]->selectionSet))
         {
             self::$args = $args;
+            $requestedFields = $info->getFieldSelection(5);
+            $parentTypeName = get_object_vars($parentType)['config']['name'];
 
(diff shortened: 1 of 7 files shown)


a70537952:
`custom_paginators` is more suitable for people that add different data into pagination depends on different type (User_Pagination, ...etc)

`pagination type` should give full control to those who need custom pagination type (Global add extra field instead of use  `custom_paginators` for every type, or change the structure of pagination)

For my usage, I currently migrating from **Folkloreatelier/laravel-graphql** package, so I need to customize data return structure

Result structure I need:
```
{
 "user": {
   "items": [],
   "cursor": {
     "total": 0, // pagination info
   }
 }
}
```
Default structure
```
{
 "user": {
   "data": [],
   "total": 0 // pagination info
 }
}
```
So I not only need to add a `cursor`, but also change `data `to `items `

src/config/config.php
```diff
@@ -164,6 +164,12 @@
         // 'my_custom_pagination' => \Path\To\Your\CustomPagination::class,
     ],
 
+    /*
+     * You can define your own pagination type.
+     * Reference \Rebing\GraphQL\Support\PaginationType::class
+     */
+    //'custom_pagination_type' => \Rebing\GraphQL\Support\PaginationType::class,
```
rebing: I think you should add a `pagination_type` instead that would default to the current `\Rebing\GraphQL\Support\PaginationType::class`

src/Rebing/GraphQL/Support/SelectFields.php
```diff
@@ -37,8 +37,24 @@ public function __construct(ResolveInfo $info, $parentType, array $args)
         if( ! is_null($info->fieldNodes[0]->selectionSet))
         {
             self::$args = $args;
+            $requestedFields = $info->getFieldSelection(5);
+            $parentTypeName = get_object_vars($parentType)['config']['name'];
+
+            // If use pagination,
+            // $parentType & $requestedFields should use dataType
+            // instead of paginationType
+            if(stripos($parentTypeName, 'pagination')){
```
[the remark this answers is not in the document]
a70537952: Updated

src/config/config.php
a70537952: Updated

src/Rebing/GraphQL/Support/SelectFields.php
```diff
@@ -37,8 +38,25 @@ public function __construct(ResolveInfo $info, $parentType, array $args)
         if( ! is_null($info->fieldNodes[0]->selectionSet))
         {
             self::$args = $args;
+            $requestedFields = $info->getFieldSelection(5);
+            $parentTypeName = get_object_vars($parentType)['config']['name'];
```
rebing: Where do you use that?

src/Rebing/GraphQL/Support/SelectFields.php
```diff
@@ -37,8 +38,25 @@ public function __construct(ResolveInfo $info, $parentType, array $args)
...
+
+            // If use pagination,
+            // $parentType & $requestedFields should use QueryType
+            // instead of paginationType
+            $paginationType = config('graphql.pagination_type', PaginationType::class);
```
rebing: This whole section looks overly complicated. For example, why do you use `get_object_vars($parentType)...` instead of just `$parentType->config ...`? It also seems like there are too many assumptions that some array keys will always exist

src/Rebing/GraphQL/Support/UploadType.php
```diff
@@ -68,4 +63,12 @@ public function parseLiteral($valueNode, ?array $variables = NULL)
     {
         throw new Error('`Upload` cannot be hardcoded in query, be sure to conform to GraphQL multipart request specification. Instead got: ' . $valueNode->kind, [$valueNode]);
     }
+
+    public static function getInstance() {
+        static $inst = null;
+        if ($inst === null) {
```
rebing: Why is that check here? If you set `static $inst = null` then it will always be null...

a70537952: Singleton design pattern, prevent creating multi instance of UploadType, if not will cause error `Schema must contain unique named types but contains multiple types named "Upload" (see http://webonyx.github.io/graphql-php/type-system/#type-registry)`, thats why we need to give a unique name for every uploadType and now we don't

src/Rebing/GraphQL/Support/SelectFields.php
a70537952: updated, also change the way to get QueryType

src/Rebing/GraphQL/Support/SelectFields.php
```diff
@@ -37,8 +38,22 @@ public function __construct(ResolveInfo $info, $parentType, array $args)
...
         {
             self::$args = $args;
+            $requestedFields = $info->getFieldSelection(5);
+
+            // If use pagination,
+            // $parentType & $requestedFields should use QueryType
+            // instead of paginationType
+            $paginationType = config('graphql.pagination_type', PaginationType::class);
+            if($info->returnType instanceof $paginationType){
+                foreach ($parentType->config['fields'] as $key => $field) {
```
rebing: I don't think that works correctly, if the pagination also has other fields that are lists. For multiple fields the last type will always be selected

a70537952: I was thinking the data with pagination will always be the last one, if there is other fields that are lists most probably these fields are customPaginator fields and selectFields dont really need to care about these fields, as long as the paginate data be the last, then we can get the correct Type

a70537952: I change to another way to achieve this, should be OK right now

PR #175, opened 2019-01-21
Take only foreign key column to avoid duplicate table prefix

### LINKED ISSUE #174: $fields->getSelect() returns `table.table.column`
Hello, I get an SQL error due to duplicate table name in qualified foreign key column:

I simplified code to get the root cause, here is my inspection:
```php
public function resolve($root, $args, SelectFields $fields)
{
    return $fields->getSelect();
}
```

It gives me:
```json
[
  "games.id",
  "games.name",
  "games.games.owner_id"
]
```

While I expected:
```json
[
  "games.id",
  "games.name",
  "games.owner_id"
]
```

To get the SQL query working.

Here is what happens inside **SelectFields.php**:

```php
$relation = call_user_func([app($parentType->config['model']), $key]);
// Add the foreign key here, if it's a 'belongsTo'/'belongsToMany' relation
if(method_exists($relation, 'getForeignKey'))
{
    $foreignKey = $relation->getForeignKey();
}
else if(method_exists($relation, 'getQualifiedForeignPivotKeyName'))
{
    $foreignKey = $relation->getQualifiedForeignPivotKeyName();
    // belongsToMany make PHP go there and $foreignKey is games.owner_id, cf getQualifiedForeignPivotKeyName() contents
}
else
{
    $foreignKey = $relation->getQualifiedForeignKeyName();
}

$foreignKey = $parentTable ? ($parentTable . '.' . $foreignKey) : $foreignKey;
// Then it prepend again the table name: games.games.owner_id
```

It seems to me like it should prepent $parentTable only if not getting a qualified name.

kylekatarnls:
Fix #174

### DIFF
--- a/src/Rebing/GraphQL/Support/SelectFields.php
+++ b/src/Rebing/GraphQL/Support/SelectFields.php
@@ -168,6 +168,8 @@ class SelectFields {
                             $foreignKey = $relation->getQualifiedForeignKeyName();
                         }
 
+                        $segments = explode('.', $foreignKey);
+                        $foreignKey = end($segments);
                         $foreignKey = $parentTable ? ($parentTable . '.' . $foreignKey) : $foreignKey;
 
                         if(is_a($relation, MorphTo::class))
--- a/src/Rebing/GraphQL/Support/Type.php
+++ b/src/Rebing/GraphQL/Support/Type.php
@@ -123,7 +123,8 @@ class Type extends Fluent {
         {
             return new InputObjectType($this->toArray());
         }
-        if ($this->enumObject) {
+        if ($this->enumObject)
+        {
             return new EnumType($this->toArray());
         }
         return new ObjectType($this->toArray());


kylekatarnls:
I did not add `[2]`.

```
$segments = explode('.', $foreignKey);
$foreignKey = end($segments);
```
is what I added and what `getForeignKey()` contains (Laravel source code). So it would be logical to use this pattern everywhere indeed.

pushchris:
@kylekatarnls, I should've commented a little better, the code I linked to is later on in the file. It's almost identical to what you changed, but was only being used for certain types of relations. Now that your change is in there, it makes the later code be missing an index and cause those relations to throw an error.

kylekatarnls:
If you can add a unit test for your error, I will try to fix this test keeping duplicate-table fix.

kylekatarnls:
I don't understand how the code chunk you show could even work, since `$foreignKey` is an array (explode), it cannot be used as a array-key.

kylekatarnls:
I will test it with my case but I think it's right and you can submit it as a pull-request.

PR #176, opened 2019-01-22
SelectFields foreign key fix

rebing:


pushchris:
@rebing does not look like this fully resolves the issue.  I am now getting errors that look something like the following: 
```
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'relation.model.model_id'
```

The foreign key in this case should be ```model.model_id```

Where the query is similar to:
```
model {
    data {
        id
        first_name
        last_name
        relation {
            id
            field
        }
    }
    total
}
```

rebing:
@pushchris Any chance you could look into the problem please and find a solution? I cannot currently test it out myself

kylekatarnls:
OK, I have an idea and will try to make the PR, but it would be great to have some unit tests for both cases (a Morph and a BelongsTo) so I could be sure not to break them when implementing it.

kylekatarnls:
Hi @rebing and @pushchris this code works for my different relations, can you check it works for you too? https://github.com/rebing/graphql-laravel/pull/177

chrispage1:
I just updated from v1.17.3 to v1.17.5.

I have an `Asset` type which has a `bookings` property with a type of `Type::listOf(\GraphQL::type('bookings'))`

My `Asset` model then has a relationship to bookings via a `HasMany` relationship

So the following - 

`query asset ($id: Int!) { asset (id: $id) { id name bookings { id } }`

would previously establish a bookings relationship with the following selects - 

- bookings.id
- bookings.asset_id

Since updating it was trying to establish a bookings relationship with these selects - 

- bookings.id
- bookings.asset.asset_id

Which of course failed and broke code. Hope you can fix :)

pushchris:
@chrispage1 check the merged pull request #177 which should resolve the issue. Having additional confirmation that this fixes it for you as well would be awesome. From what I can tell this hasn't been pushed to a release yet since 1.8.0 was pulled.

kylekatarnls:
Yep, you would need to test with the version `dev-master` and it could be good to wait for your confirmation before creating the tag :)

chrispage1:
@pushchris @kylekatarnls - it's a project in development so no harm. I'll give it a go today.

chrispage1:
Just upgraded to `v1.17.6` and all is working now. Thanks!

lucasantoniooficial:
I have this problem in package version 8.6, how to solve it?

lucasantoniooficial:
it works only for id column, when I get the other columns it gives an undefined column error

PR #230, opened 2019-03-05
Fix for infinite recursion when InputTypeObject contains self reference fields

infinity0n3:
A simple Input object with a self reference field causes the graphql library to end up in an uncontrolled recursive loop between `Field->inferRulesFromType` and `Field->getInputTypeRules` trying to generate the field rules.

```
...
class RepositoryFilterInput extends GraphQLType
{
    protected $inputObject = true;
    
    protected $attributes = [
        'name' => 'RepositoryFilter',
        'description' => 'A type RepositoryFilter',
    ];

    public function fields()
    {
        return [

            'parent' => [
                'type' => GraphQL::type('RepositoryFilter'),
                'description' => 'The parent of RepositoryFilter'
            ],
       ]
   }
...

```
An uncontrolled recursive call is extremely dangerous as it will bring down the php interpreter (and possibly more, depending on how much RAM and CPUs you have).

The proposed fix is to simply skip the recursive call for self referenced fields. I've tested it with my full implementation of filters and it works even with nested ones for parent without any issues.

### DIFF
--- a/src/Rebing/GraphQL/Support/Field.php
+++ b/src/Rebing/GraphQL/Support/Field.php
@@ -135,6 +135,12 @@ class Field extends Fluent {
 
             // then recursively call the parent method to see if this is an
             // input object, passing in the new prefix
+            if($field->type instanceof InputObjectType) {
+                // in case the field is a self reference we must not do
+                // a recursive call as it will never stop
+                if($field->type->toString() == $input->toString())
+                    continue;
+            }
             $rules = array_merge($rules, $this->inferRulesFromType($field->type, $key, $resolutionArguments));
         }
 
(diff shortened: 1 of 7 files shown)


mfn:
Please add a test for this!

infinity0n3:
Should I extend the example code with Repository and RepositoryInput so that those can be used in the tests?

infinity0n3:
I've added the test case that fails when the fix is not available and works when it is applied.
The test fails in an interesting way as it only returns a `null` response. I guess when running phpunit tests php expects bad stuff so it handles the recursion limit better and stops at 256 iterations.

Now this is only visible when one uses XDebug to capture the exceptions as only then at some point that 256 limit is visible.

Never the less. The test case is here and proves the fix works.

PR #250, opened 2019-04-24
Use getter to retrieve types

dnyg:
I want to override the `getTypes()` function to filter some of the types in retrospection queries for my own project.

Currently overwriting the `->getTypes()` function does nothing, because it is not used.


rebing:
Probably should replace other instances of `$this->types` as well?

PR #253, opened 2019-05-15
Support mongodb

WilliamPeralta:
mongodb not support table name as prefix of field name.


rebing:
That is definitely not the correct solution, because any table can have `<any>_id` as the primary key

WilliamPeralta:
@rebing sorry I made a solution on the fly .. anyway I sent a pull request with an improved solution

PR #262, opened 2019-05-19
Add fields to the default pagination type

ZaidBarghouthi:
Added last_page (number of pages) and has_more_pages fields.

I believe these fields are quite useful to have.

The description might be improved though! 

Cheers.

### DIFF
--- a/src/Rebing/GraphQL/Support/PaginationType.php
+++ b/src/Rebing/GraphQL/Support/PaginationType.php
@@ -60,13 +60,13 @@ class PaginationType extends ObjectType {
                 'selectable'    => false,
             ],
             'last_page' => [
-                'type'          => GraphQLType::int(),
+                'type'          => GraphQLType::nonNull(GraphQLType::int()),
                 'description'   => 'The last page (number of pages)',
                 'resolve'       => function(LengthAwarePaginator $data) { return $data->lastPage(); },
                 'selectable'    => false,
             ],
             'has_more_pages' => [
-                'type'          => GraphQLType::boolean(),
+                'type'          => GraphQLType::nonNull(GraphQLType::boolean()),
                 'description'   => 'Determines if cursor has more pages after the current page',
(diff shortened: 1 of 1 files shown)


src/Rebing/GraphQL/Support/PaginationType.php
```diff
@@ -59,6 +59,18 @@ protected function getPaginationFields($typeName)
                 'resolve'       => function(LengthAwarePaginator $data) { return $data->lastItem(); },
                 'selectable'    => false,
             ],
+            'last_page' => [
+                'type'          => GraphQLType::int(),
```
mfn: This would be wrapped in `GraphQLType::nonNull()` as the the returned value of `lastPage()` is always defined as int

src/Rebing/GraphQL/Support/PaginationType.php
```diff
@@ -59,6 +59,18 @@ protected function getPaginationFields($typeName)
...
+                'description'   => 'The last page (number of pages)',
+                'resolve'       => function(LengthAwarePaginator $data) { return $data->lastPage(); },
+                'selectable'    => false,
+            ],
+            'has_more_pages' => [
+                'type'          => GraphQLType::boolean(),
```
mfn: This also needs to be wrapped in `GraphQLType::nonNull()`, `hasMorePages` _always_ returns a boolean

src/Rebing/GraphQL/Support/PaginationType.php
ZaidBarghouthi: Ok

src/Rebing/GraphQL/Support/PaginationType.php
ZaidBarghouthi: Ok :)

ZaidBarghouthi:
Is there a test file for this?

Issue #254, opened 2019-05-15
array_get is deprecated and will be removed in Laravel 5.9


JakeBooher:
Not sure if this is the only similar issue, but just found this while replacing the validator `Rebing\GraphQL\Support\Field`  with our custom version while attempting to upgrade from the old Folklore extension. All of the array_* helpers have been replaced with the `Illuminate\Support\Arr` util class

mfn:
See https://github.com/rebing/graphql-laravel/pull/263

rebing:
#263

PR #263, opened 2019-05-19
Replace array/str global helpers with class

mfn:
Some of the global helpers, including `array_*` and `str_*` are being deprecated in the future of Laravel.

They will always be available via a third party package for _Applications_, but libraries should not depend upon them.

See https://github.com/laravel/framework/pull/27504 and https://github.com/illuminate/support/commit/a41d7e12b39edea740d70a5591d3239284b6d16e#diff-37103b40d038920dbb5e8b957d21a0c8


mfn:
Fixes https://github.com/rebing/graphql-laravel/issues/254

Issue #259, opened 2019-05-18
Dedicate names for routes


mfn:
A useful feature of Folklore I miss in Rebing are the explicit names of routes.

Relevant parts from Folklore (look for the `'as'` option:
```php
            $router->get(preg_replace($schemaParameterPattern, '', $queryRoute), array(
                'as' => 'graphql.query',
                'uses' => $queryController
            ));
            $router->get(preg_replace($schemaParameterPattern, '{graphql_schema}', $queryRoute), array(
                'as' => 'graphql.query.with_schema',
                'uses' => $queryController
            ));
            $router->post(preg_replace($schemaParameterPattern, '', $queryRoute), array(
                'as' => 'graphql.query.post',
                'uses' => $queryController
            ));
```
In artisan:
```php
./artisan route:list|grep -i graphql
|        | GET|HEAD                               | graphiql/{graphql_schema?}                            | graphql.graphiql
|        | POST                                   | graphql/{graphql_schema?}                             | graphql.query.post
|        | GET|HEAD                               | graphql/{graphql_schema?}                             | graphql.query
```
This is also useful in tests and other parts of the code where you want to get the URL of the route. You just call `route('graphql.query.post')` and have a proper URL to.

In Rebing the routes are defined in `vendor/rebing/graphql-laravel/src/Rebing/GraphQL/routes.php` but no `'as'` is used to explicitly name them:
```
$ ./artisan route:list|grep -i graphql
|        | GET|HEAD                               | graphql                                               |                                 | Rebing\GraphQL\GraphQLController@query
|        | POST                                   | graphql                                               |                                 | Rebing\GraphQL\GraphQLController@query
|        | POST                                   | graphql/{default}                                     |                                 | Rebing\GraphQL\GraphQLController@query
|        | GET|HEAD                               | graphql/{default}                                     |                                 | Rebing\GraphQL\GraphQLController@query
```
Would you consider them?

mfn:
(ps: also for graphiql)

rebing:
Yes, you can create a PR :)

PR #264, opened 2019-05-20
Add names to all routes

mfn:
- Convenient when you want to link to them, e.g. within integration tests
- For convenience, and Folklore compat, remove the `get` in the name
  from `GET` routes
- Don't define mutation route if it's the same as the query route.
  Otherwise also the route names would overwrite each other and having
  a default route named `graphql.query` feels more idiomatic, regardless
  of whether it's a mutation or not (unless they would be truly different
  routes)

Example with Laravels `route:list`:
```
$ ./artisan route:list|grep -i graphql
|        | POST                                   | api/graphiql/{graphql_schema?}                            | graphiql.post                   | Rebing\GraphQL\GraphQLController@graphiql                         | web-graphiql                                         |
|        | GET|HEAD                               | api/graphiql/{graphql_schema?}                            | graphiql                        | Rebing\GraphQL\GraphQLController@graphiql                         | web-graphiql                                         |
|        | POST                                   | api/graphiql/{graphql_schema?}/{default}                  | graphiql.default.post           | Rebing\GraphQL\GraphQLController@graphiql                         | web-graphiql                                         |
|        | GET|HEAD                               | api/graphiql/{graphql_schema?}/{default}                  | graphiql.default                | Rebing\GraphQL\GraphQLController@graphiql                         | web-graphiql                                         |
|        | GET|HEAD                               | api/graphql                                               | graphql.query                   | Rebing\GraphQL\GraphQLController@query                            | web-graphql                                          |
|        | POST                                   | api/graphql                                               | graphql.query.post              | Rebing\GraphQL\GraphQLController@query                            | web-graphql                                          |
|        | POST                                   | api/graphql/{default}                                     | graphql.default.post            | Rebing\GraphQL\GraphQLController@query                            | web-graphql                                          |
|        | GET|HEAD                               | api/graphql/{default}                                     | graphql.default                 | Rebing\GraphQL\GraphQLController@query                            | web-graphql                                          |
```

Fixes https://github.com/rebing/graphql-laravel/issues/259

### DIFF
--- a/src/Rebing/GraphQL/routes.php
+++ b/src/Rebing/GraphQL/routes.php
@@ -50,11 +50,16 @@ $router->group(array_merge([
             $defaultMethod = config('graphql.schemas.' . config('graphql.default_schema') . '.method', ['get', 'post']);
 
             foreach ($defaultMethod as $method) {
+                $routeName = 'graphql.query';
+                if($method !== 'get') {
+                    $routeName .= ".$method";
+                }
                 $router->{$method}(
                     preg_replace($schemaParameterPattern, '', $queryRoute),
                     [
                         'uses'          => $queryController,
                         'middleware'    => $defaultMiddleware,
+                        'as'            => $routeName,
                     ]
                 );
             }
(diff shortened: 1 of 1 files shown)


src/Rebing/GraphQL/routes.php
```diff
@@ -78,37 +88,53 @@
...
+                'uses' => $queryController,
+                'as'   => 'graphql.query.post',
+            ]);
         }
     }
 
-    // Mutation
-    if($mutationRoute)
+    // Mutation routes (define only if different than query)
+    if($mutationRoute && $mutationRoute !== $queryRoute)
```
mfn: This change is unrelated to the PR but using named routes now made it obvious (well, at least to me) that there's no need to define the mutation routes explicitly if they're in fact the same as the query routes.

Otherwise the route names override each other we will _always_ end up with `graphql.mutation.*` route names, no matter what 🤷‍♀️

Issue #258, opened 2019-05-18
Custom context overwritten with SelectFields class


mfn:
Trying to migrate from Folklore I face the problem that I'm using a custom `Context` being passed as the 3rd argument which is overwritten in `\Rebing\GraphQL\Support\Field::getResolver`.

At this point `$arguments[2]` is my custom Context class I need for the request:
```php
        return function() use ($resolver, $authorize)
        {
            $arguments = func_get_args();
```
but a few lines later, this happens:
```php
            // Replace the context argument with 'selects and relations'
            // $arguments[1] is direct args given with the query
            // $arguments[2] is context (params given with the query)
            // $arguments[3] is ResolveInfo
            if(isset($arguments[3]))
            {
                $fields = new SelectFields($arguments[3], $this->type(), $arguments[1]);
                $arguments[2] = $fields;
            }
```
The comment even mentions the context, but it overwrites it.

I don't mind the `SelectFields` approach and acknowledge it's usefulness, but every (also non-PHP) server-side GraphQL implementation I've seen so far uses the same argument pattern "root, args, context, resolver info", thus overwriting this seems a bit odd.

What is the recommended way to have the context available, too?

Should I override `\Rebing\GraphQL\Support\Field::getResolver`?

rebing:
I agree that it should not be replaced in such a way and SelectFields should instead be the 5th argument. If you can override the `getResolver` then it's probably your best option right now

mfn:
@rebing would you be open to fix this, considering it's a breaking change, for a possible `2.*` release?

PR #274, opened 2019-05-21
[BREAKING CHANGE 🚨] Append SelectFields support class instead of overriding the context, make it deferred

mfn:
# This is a hard breaking change
This PR is to start a discussion / viability / willingness for this change.

### Background
Most GraphQL server implementations follow the following pattern when calling query/mutation resolvers: root, args, context, info

This library opted to overwrite context with the useful "SelectFields" class.

However, overwriting means the context isn't available anymore at all. Which is quite a surprise as there are useful application making use of the context.

Whilst not the most useful thing, the default context for example is the authenticated user, if any is present. This is even implemented by this library in `\Rebing\GraphQL\GraphQLController::queryContext`:
```php
    protected function queryContext()
    {
        try {
            return app('auth')->user();
        } catch (\Exception $e) {
            return;
        }
    }
```
Given the design of the library, `queryContext` can be overridden by specifying a custom `GraphQLController` in the configuration. However this is moot as the context is never passed to any query/mutation.

Other library examples:
- Apollo graphql-tools https://www.apollographql.com/docs/graphql-tools/resolvers
  > `author(obj, args, context, info) {`
- The definition in the underlying webonxy PHP library http://webonyx.github.io/graphql-php/executing-queries/
  > Any value that holds information shared between all field resolvers. Most often they use it to pass currently logged in user, locale details, etc.
  >
  > It will be available as the 3rd argument in all field resolvers. (see section on Field Definitions for reference) graphql-php never modifies this value and passes it as is to all underlying resolvers.

## Further change
Besides re-ordering the arguments I've also changed `SelectFields $selectFields` to a `Closure $getSelectField`.

The reason is that the use of `SelectFields` is entirely optional, yet as now to provide this parameter, it has to traverse the GraphQL AST to extract the necessary information.

In cases where this is not used, it's waste of resources.

That's why the Closure was added which lazy creates SelectFields only when requested.

### Migration
Every query/mutation typing more than the first two arguments to `resolve` needs to adapt them (types are **not** affected):
```php
public function resolve($root, $args, SelectFields $fields, ResolveInfo $resolveInfo)
```
to
```php
public function resolve($root, $args, $context, ResolveInfo $resolveInfo, Closure $getSelectFields) {
  $selectFields = $getSelectFields();
}
```
### Compatibility
This change is inherently not backwards compatible. Although unclear if this library follows [SemVer](https://semver.org/), such a change should lead to a new major version bump (i.e. `v2.*` as it currently stands).

### TODO
- [x] There's currently no test making actual use of the `SelectFields` utility => one should be added
  => Fixed after https://github.com/rebing/graphql-laravel/pull/289 landed which added some
- [ ] ~Add dedicated entry to the Readme to inform users~
  Decided against it, the `CHANGELOG.md` will have the details
- [x] If https://github.com/rebing/graphql-laravel/pull/273 gets merged, also add a clear entry about it

### Links
- Fixes https://github.com/rebing/graphql-laravel/issues/258

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,9 @@ CHANGELOG
 [Next release](https://github.com/rebing/graphql-laravel/compare/v1.23.0...master)
 ------------
 ## Breaking changes
-- The order and arguments for resolvers has changed:
-  - before: `resolve($root, $array, SelectFields $fields, ResolveInfo $info)`
-  - after: `resolve($root, $array, $context, ResolveInfo $info, SelectFields $fields)`
+- The order and arguments/types for resolvers has changed:
+  - before: `resolve($root, $array, SelectFields $selectFields, ResolveInfo $info)`
+  - after: `resolve($root, $array, $context, ResolveInfo $info, Closure $getSelectFields)`
 - Added PHP types / phpdoc to all methods / properties [\#331](https://github.com/rebing/graphql-laravel/pull/331)
   - Changes in method signatures will require small adaptions.
 - Validation errors are moved from error.validation to error.extensions.validation as per GraphQL spec recommendation [\#294](https://github.com/rebing/graphql-laravel/pull/294)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ CHANGELOG
 [Next release](https://github.com/rebing/graphql-laravel/compare/v1.23.0...master)
 ------------
 ## Breaking changes
+- The order and arguments for resolvers has changed:
+  - before: `resolve($root, $array, SelectFields $fields, ResolveInfo $info)`
+  - after: `resolve($root, $array, $context, ResolveInfo $info, SelectFields $fields)`
 - Added PHP types / phpdoc to all methods / properties [\#331](https://github.com/rebing/graphql-laravel/pull/331)
   - Changes in method signatures will require small adaptions.
 - Validation errors are moved from error.validation to error.extensions.validation as per GraphQL spec recommendation [\#294](https://github.com/rebing/graphql-laravel/pull/294)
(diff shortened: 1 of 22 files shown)


mfn:
Note: after https://github.com/rebing/graphql-laravel/issues/286 this means merging this won't affect the still available `v1` branch and release line.

rebing:
I agree that the context field should not be overridden. However, perhaps there is an easier way to retrieve the SelectFields results as always writing out all 5 function arguments is not convenient. An ideal solution would be to just call some parent function, e.g
```
public function resolve ()
{
  $with = $this->getRelations();
  $columns = $this->getSelect();
  ...
}
```
but if this isn't possible then we have to go with adding the fields as the 5th argument. Either way it will be a breaking change.

mfn:
Thanks for bringing this up.

I'm not particular fond of adding such methods to the (probably) `Field` class. The nice thing about the current approach is that, through the encapsulation, it's easy to understand why you can call these two methods on the `SelectFields` object => this would change if they were just more methods on `Field`.

After consider your idea and reading the code a bit more, I realized something else: SelectFields is always created, always instantiated, always processed.

Even if the user doesn't use it.

IMHO: as useful as SelectFields is, it should be an optional utility without overhead if not used.

I will ponder more about an approach which:
- does not obscure the methods passed to the resolver
- does not incur overhead if not used

mfn:
Added blocking label until there's better plan.

Last change is that I implemented lazy instantiation of `SelectFields`.

mfn:
@rebing I did let some time pass but couldn't come up with something I would deem better and would like to move forward here.

I've updated PR description, but the TL;DR is:
- move SelectFields to last query/mutation resolve argument (5th position)
- make it lazy instantiated via a Closure to not generate it at all times when it's maybe even not used

mfn:
Thanks!

I think not long after this is merged we should cut a release candidate for the `v2` line to gather feedback on all the changes that went into it.

PR #277, opened 2019-05-23
[BREAKING CHANGES] A Type always has it's Interface fields

### LINKED ISSUE #272: Interfaces and SelectFields
Hi,

I'm trying to request a sum of a column, and a group by column, so MySQL request should be really simple, something like 

```sql
select sum(revenue) as revenue, attributed_at from revenues group by attributed_at;
```

Simple, right ? Well, now I have multiple tables containing (other) revenues, but structured the same way. The best solution I found was using an Interface, since most of the structure is the same. So I did it this way:

```php
<?php

namespace App\GraphQL\Interfaces;

use GraphQL\Type\Definition\Type;
use Illuminate\Database\Query\Builder;
use Illuminate\Support\Facades\DB;
use Rebing\GraphQL\Support\InterfaceType;

class RevenueInterfaceType extends InterfaceType
{

    /**
     * @var array
     */
    protected $attributes = [
        'name' => 'Revenue',
        'description' => 'A revenue',
    ];

    /**
     * @return array
     */
    public function fields(): array
    {
        return [
            'attributed_at' => [
                'description' => 'The id of the revenue'
            ],
            'revenue' => [
                'type' => Type::nonNull(Type::float()),
                'description' => 'The revenues',
                'alias' => DB::raw('sum(revenue) as revenue'),
            ],
        ];
    }
}
```

Now, I'm making my request like so:

```graphql
query Revenues{
  revenues{
    revenue
    attributed_at
  }
}
```

But when I'm using the `SelectFields::getSelect()` in my query, it returns `['*']`.

After digging into the `SelectFields` class, I found that, if the `$parentType` is an instance of `InterfaceType`, it just returns `['*']` (a test that could've been done at the very top of the method, by the way).

In my point of view, this is not a good way of handling it, since `alias` is not used, and most of all, the `interface` should be a manner of saying: "_This field is available for this type, no matter what_". Maybe I'm wrong though and not understanding the way it should work. If so, would anybody help me figure out how i could achieve this ?

Thanks

EdwinDayot:
### Background

Query's type defined as an interface leads to a problem: the `SelectFields::getSelect` returns: `['*']`.

The code responsible of this is stating the following:

```php
/* SelectFields.php > handleFields */

// If parent type is an interface or union we select all fields
// because we don't know which other fields are required
// from types which implement this interface
```

The reason why this code was implemented seems to be a specific case. If other fields are required, they probably should be added to the current selected fields by the developer in it's `Query` component.

GraphQL.org is stating that
> Like many type systems, GraphQL supports interfaces. An Interface is an abstract type that includes a certain set of fields that a type must include to implement the interface.
[https://graphql.org/learn/schema/#interfaces](https://graphql.org/learn/schema/#interfaces)

Since we know what fields are required, we can safely select the fields of the `Interface`.

This behavior leads to "1.5" issues:
* The `alias` is not used to retrieve fields, so it can lead to unexpected issues.
* When using `group by`, MySQL with `sql_mode=ONLY_FULL_GROUP_BY` will throw an error since all fields aren't in the group by. (Since it is related to a specific configuration of a specific database system, this can be ignored)

### Impact

The `Interface`s will need to explicitly define their available fields, or be selected by the developer in their `Query` component.

### Compatibility

This change will only impact the `Queries` that requires more fields than the actual `Interface` provides. Since it's a "rollback", it's not backward compatible as it could cause bugs in currently deployed codes.

### TODO

- [ ] Write a test suite

### Links

* Fixes #272

### DIFF
--- a/src/Rebing/GraphQL/Support/SelectFields.php
+++ b/src/Rebing/GraphQL/Support/SelectFields.php
@@ -213,7 +213,7 @@ class SelectFields
             }
         }
 
-        // If parent type is an or union we select all fields
+        // If parent type is an union we select all fields
         // because we don't know which other fields are required
         if (is_a($parentType, UnionType::class)) {
             $select = ['*'];
--- a/src/Rebing/GraphQL/Support/SelectFields.php
+++ b/src/Rebing/GraphQL/Support/SelectFields.php
@@ -8,7 +8,6 @@ use GraphQL\Error\InvariantViolation;
 use GraphQL\Type\Definition\UnionType;
 use GraphQL\Type\Definition\ListOfType;
 use GraphQL\Type\Definition\ResolveInfo;
-use GraphQL\Type\Definition\InterfaceType;
 use Illuminate\Database\Eloquent\Relations\HasOne;
 use Illuminate\Database\Eloquent\Relations\HasMany;
 use Illuminate\Database\Eloquent\Relations\MorphTo;
@@ -214,9 +213,8 @@ class SelectFields
             }
         }
 
-        // If parent type is an interface or union we select all fields
+        // If parent type is an or union we select all fields
         // because we don't know which other fields are required
-        // from types which implement this interface
         if (is_a($parentType, UnionType::class)) {
             $select = ['*'];
         }
--- a/src/Rebing/GraphQL/Support/SelectFields.php
+++ b/src/Rebing/GraphQL/Support/SelectFields.php
@@ -217,7 +217,7 @@ class SelectFields
         // If parent type is an interface or union we select all fields
         // because we don't know which other fields are required
         // from types which implement this interface
-        if (is_a($parentType, InterfaceType::class) || is_a($parentType, UnionType::class)) {
+        if (is_a($parentType, UnionType::class)) {
             $select = ['*'];
         }
     }


mfn:
> Write a test suite

This is also my biggest concern right now :-)

In a perfect world:
- a test would be written against master
- then this PR is rebased and then tests show the change

But I understand this is an issue, as there's currently no infrastructure for integration tests sadly.

mfn:
Label as blocked until https://github.com/rebing/graphql-laravel/pull/289 can land, after which we can cover the change with a test

mfn:
@EdwinDayot I want to tackle this next.

I think on a high level I understand the issue, but it would help me if you could give a more specific example how to set this up. As I mentioned in https://github.com/rebing/graphql-laravel/pull/277#issuecomment-496212957, I would prefer to make a tests in master first so this PR will then also show it's change to it.

The issue you referenced talks about aggregation functions and aliases too. Are they both really relevant? Or it it maybe only the alias?

EdwinDayot:
Sure, my use case was having a RevenueInterface like this:
```php
<?php

namespace App\GraphQL\Type;

use GraphQL\Type\Definition\Type;
use Illuminate\Database\Query\Builder;
use Illuminate\Support\Facades\DB;
use Rebing\GraphQL\Support\Facades\GraphQL;
use Rebing\GraphQL\Support\InterfaceType;

class RevenueInterfaceType extends InterfaceType
{
    ...

    /**
     * @return array
     */
    public function fields(): array
    {
        return [
            'revenues' => [
                'type' => Type::nonNull(Type::float()),
                'description' => 'The revenues',
                'alias' => DB::raw('sum(revenues) as revenues'),
            ],
            'attributed_at' => [
                'description' => 'The date the revenue has been attributed at'
            ],
        ];
    }

    ...
}
```

And 2 or more Types implementing this interface.

The aggregating functions aren't relevant here.

mfn:
Should have changed base branch before pushing the merge 🤷‍♀️

Anyway, merged via https://github.com/rebing/graphql-laravel/commit/bea5ec7a7a5363ed0bb90d9198830e41981457f7

Issue #280, opened 2019-05-26
Alias is not working with me! 


ZaidBarghouthi:
The following is not working with me. It might be doing something wrong, I am not absolutely certain. I'm getting `null` when querying the field. 

```php
public function fields()
{
    return [
        ...
        'createdAt' => [ 
            'type' => Type::string(),
            'description' => 'The created at timestamp',
            'alias' => 'created_at', 
        ],
        ...
    ];
}
```

but the following is working fine
```php
public function fields()
{
    return [
        ...
        'createdAt' => [ 
            'type' => Type::string(),
            'description' => 'The created at timestamp',
            'resolve' => function($type) { return $type->created_at; }, 
        ],
        ...
    ];
}
```

Am I missing something?

EdwinDayot:
Hi,

Did you try the following ?

```php
public function fields()
{
    return [
        ...
        'createdAt' => [ 
            'type' => Type::string(),
            'description' => 'The created at timestamp',
            'alias' => 'created_at',
            'resolve' => function($type) { return $type->created_at; }, 
        ],
        ...
    ];
}
```

The `alias` is necessary for the query (for instance: `select created_at from ...`), the `resolve`function gets the aliased field from the Eloquent instance.

mfn:
I guess the case could be made that when providing the `alias`, the default behaviour would be to also return it (unless a custom resolver like above is provided)?

EdwinDayot:
Since it can be confusing, it would probably be a good idea

ZaidBarghouthi:
> Hi,
> 
> Did you try the following ?
> 
> ```
> public function fields()
> {
>     return [
>         ...
>         'createdAt' => [ 
>             'type' => Type::string(),
>             'description' => 'The created at timestamp',
>             'alias' => 'created_at',
>             'resolve' => function($type) { return $type->created_at; }, 
>         ],
>         ...
>     ];
> }
> ```
> 
> The `alias` is necessary for the query (for instance: `select created_at from ...`), the `resolve`function gets the aliased field from the Eloquent instance.

Yes. It works, but its works just as well without `alias`!. So I'm not sure what is the proper use for alias!

EdwinDayot:
Are you selecting specific fields in your `Query`? Or are you selecting `*`?

ZaidBarghouthi:
not sure how the following is translated into SQL >>
```gql
{
  userById(id: 17) {
    id
    audits(limit:4){
      data{
        id
        createdAt // here <----
[...]

PR #283, opened 2019-05-28
Auto-resolve unresolved aliased fields

ZaidBarghouthi:
Adds simple default resolver to aliased fields when no resolver is present. 

**I'm not sure how to test this properly ( please comment )**

#280

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG
 Next release
 ------------
 - This project has a changelog `\o/`
+- Auto-resolve aliased fields [\#283](https://github.com/rebing/graphql-laravel/pull/283)
 
 2019-03-07, v1.21.2
 -------------------
--- a/src/Rebing/GraphQL/Support/Type.php
+++ b/src/Rebing/GraphQL/Support/Type.php
@@ -49,6 +49,15 @@ class Type extends Fluent
                 return call_user_func_array($resolver, $args);
             };
         }
+        
+        if (isset($field['alias'])) {
+            $alias = $field['alias'];
+
+            return function ($type) use ($alias)
+            {
+                return $type->{$alias}; 
+            };
+        }
     }
 
     public function getFields()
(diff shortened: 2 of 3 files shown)


ZaidBarghouthi:
please consider adding a `editorconfig` file.

mfn:
> please consider adding a `editorconfig` file.

👍 added in https://github.com/rebing/graphql-laravel/commit/b7ec923370e0bf3fd85ed1d83074f105ef50a828 !

ZaidBarghouthi:
Should we have a check that this new resolver only applies when there is a model defined?

mfn:
Label as blocked until https://github.com/rebing/graphql-laravel/pull/289 can land, after which we can cover the change with a test

mfn:
In https://github.com/rebing/graphql-laravel/pull/289 I've now added the test`\Rebing\GraphQL\Tests\Database\SelectFieldsTest::testWithSelectFieldsAndModelAndAlias` which I expect will "break" and needs to be adjusted after this PR.

ZaidBarghouthi:
Yup.

mfn:
Thanks for the PR!

I've merged this into master and also the [recent v1 branch](https://github.com/rebing/graphql-laravel/issues/286) and cut a [new v1.22.0 release](https://github.com/rebing/graphql-laravel/releases/tag/v1.22.0) (hopefully it works 🤞 as expected).

> Should we have a check that this new resolver only applies when there is a model defined?

There does not seem to be a technical reason for this _yet_.

For simple attribute aliases it seems the model is not required. Without it, the generated SQL query will simple not use qualified names. I.e. the difference will be:

- with model
  ```sql
  select "posts"."id", "posts"."title" from "posts" where "posts"."id" = ? limit 1;
  ```
- without model
  ```sql
  select "id", "title" from "posts" where "posts"."id" = ? limit 1;
  ```
But nevertheless it will work correctly; as the SelectFields in general does.

ZaidBarghouthi:
Thank you @mfn

PR #289, opened 2019-05-30
[TESTS] Refactor to support database tests via sqlite

mfn:
## Summary
_Only "tests" related code has been touched in this PR!_

I've already started [moving tests into a dedicated namespace](https://github.com/rebing/graphql-laravel/commit/44185e50665fd7bfe37c736789ee9a699fec827f)
- create new namespace/directory layout for tests and supporting infrastructure
  image
- Add a new `TestCaseDatabase` class which will setup an in-memory SQLITE database and see it with migrations/factories from `tests/Support/database`
  mainly, we will split tests between pure unit tests (no database required) and ones with
### TODO
- [x] Add actual tests using the database, e.g. SelectFields tests
- [ ] ~When endpoint test fails under certain error conditions, they return a regular 404 error page with the exception included in the HTML => this makes debugging harder than it should be~
  Will tackle later.

This should pave the way for https://github.com/rebing/graphql-laravel/pull/277 and https://github.com/rebing/graphql-laravel/pull/283 so we we'll have proper tests for them.


ZaidBarghouthi:
Eventually we need integration tests right?

mfn:
> Eventually we need integration tests right?

Ack, see the TODO ;-)

tests/Database/SelectFieldsTest.php
```diff
@@ -0,0 +1,115 @@
...
+GRAQPHQL;
+
+        $this->sqlCounterReset();
+
+        $response = $this->call('GET', '/graphql', [
+            'query' => $graphql,
+        ]);
+
+        $this->assertSqlQueries(<<<'SQL'
+select "posts"."id", "posts"."title" from "posts" where "posts"."id" = ? limit 1;
```
mfn: This test shows the functionality with `SelectFields` `\o/`

rebing: Nice

tests/Support/Queries/PostQuery.php
```diff
@@ -0,0 +1,38 @@
...
+            'id' => [
+                'type' => Type::nonNull(Type::id()),
+            ],
+        ];
+    }
+
+    public function resolve($root, $args, SelectFields $selectFields)
+    {
+        return Post
+            ::findOrFail($args['id']);
```
rebing: Is this kind of formatting standard?

tests/Support/Queries/PostQuery.php
```diff
@@ -0,0 +1,38 @@
...
+
+class PostQuery extends Query
+{
+    protected $attributes = [
+        'name' => 'post',
+    ];
+
+    public function type()
+    {
+        return GraphQL::type('Post');
```
rebing: Shouldn't this (and the other Queries as well) return `Type::nonNull(GraphQL::type('Post'))` instead?

tests/Database/SelectFieldsTest.php
```diff
@@ -95,6 +97,40 @@ public function testWithSelectFieldsAndModel(): void
...
+        ]);
+
+        $graphql = <<<GRAQPHQL
+{
+  postWithSelectFieldsAndModelAndAlias(id: $post->id) {
+    id
+    description
+  }
+}
+GRAQPHQL;
```
ZaidBarghouthi: Whats the issue with indenting this?

tests/Database/SelectFieldsTest.php
```diff
@@ -95,6 +97,40 @@ public function testWithSelectFieldsAndModel(): void
...
+
+        $this->sqlCounterReset();
+
+        $response = $this->call('GET', '/graphql', [
+            'query' => $graphql,
+        ]);
+
+        $this->assertSqlQueries(<<<'SQL'
+select "posts"."id", "posts"."title" from "posts" where "posts"."id" = ? limit 1;
```
ZaidBarghouthi: Whats the issue with indenting this?

tests/Support/Queries/PostQuery.php
rebing: Ok nvm, I just saw the issue you posted. It's strange though, because I thought I've used nonNull with custom types

tests/Support/Queries/PostQuery.php
mfn: Ops, probably not. I'm so used to it because _usually_ I've more conditions and then I start if off like this to get to this:
```php
return Post
  ::select()
  ->where()
  ->where()
  ->orderBy()
  ->get();
```
But yes, pointless here!

tests/Database/SelectFieldsTest.php
mfn: What do you mean specifically?

tests/Database/SelectFieldsTest.php
mfn: Can you elaborate?

tests/Support/Queries/PostQuery.php
mfn: Fixed (one line)

mfn:
> The `assertSqlQueries` solution is cool.

Thanks!

I don't see this approach used very often. I've been working on commercial projects where, under complex circumstances, it's necessary to understand the SQL code being executed (selected columns, n+1 problems, redundant queries, …).
n+1 problems can sometimes also easily be spotted via just assertion the counts by seeing the actual statements it's very insightful.

_Especially_ in this case where the entire goal of `SelectFields` is about custom queries, I though the only way to assert the correct is about the SQL statements.

tests/Database/SelectFieldsTest.php
ZaidBarghouthi: its kinda painful to work this way! My editor went mad!
[Image](https://imgur.com/PsULGAc)

img

mfn: Hmm, works correctly for me => using PhpStorm 2019 here; what IDE are you using?

collapse_expand

The issue I'm seeing:
- it doesn't matter for the GraphQL queries (whitespace is insignificant)
- but for the SQL assertion/comparison, they are significant 

Another reason I like to keep it this way is, especially for the SQL assertions, they _can_ get very long. Inevitably horizontal scrolling for them will be necessary often, but the less that is required, the better.

Issue #290, opened 2019-05-30
SelectFields doesn't work when return type is wrapped in Type::nonNull


mfn:
From upcoming tests in https://github.com/rebing/graphql-laravel/pull/289

This works:
```php
class PostQueryWithSelectFields extends Query
{
    protected $attributes = [
        'name' => 'postWithSelectFields',
    ];

    public function type()
    {
        return GraphQL::type('Post');
    }

    public function args()
    {
        return [
            'id' => [
                'type' => Type::nonNull(Type::id()),
            ],
        ];
    }

    public function resolve($root, $args, SelectFields $selectFields)
    {
        return Post
            ::select($selectFields->getSelect())
            ->findOrFail($args['id']);
    }
}
```
but when the type is changed to this:
```php
    public function type()
    {
        return Type::nonNull(GraphQL::type('Post'));
    }
```
then the code fails with:
`SQLSTATE[HY000]: General error: 1 near "from": syntax error (SQL: select  from "posts" where "posts"."id" = 1 limit 1)"`

PR #292, opened 2019-05-31
[SELECTFIELDS] Unwrap type before introspecting for fields

mfn:
## Summary
Makes queries work where they type is wrapped in nonNull for example.

By using the recursive version of `getWrappedType($recursive=true)` we also make cases work where the types are nested like this: `Type::nonNull(Type::listOf(Type::nonNull(GraphQL::type(…))));`

Fixes https://github.com/rebing/graphql-laravel/issues/290

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ Next release
 ### Changed
 - Replace global helper `is_lumen` with static class call `\Rebing\GraphQL\Helpers::isLumen`
 
+### Fixed
+- SelectFields now works with wrapped types (nonNull, listOf)
+
 2019-03-07, v1.21.2
 -------------------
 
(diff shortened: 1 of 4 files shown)


mfn:
Targeting also v1 releae

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ Next release
 ### Changed
 - Replace global helper `is_lumen` with static class call `\Rebing\GraphQL\Helpers::isLumen`
 
+### Fixed
+- SelectFields now works with wraped types (nonNull, listOf)
```
rebing: Typo `wrapped`

mfn: Fixed 👍

mfn:
I've added one more commit where I changed an existing getWrappedType check.

This PR doesn't include a test for it but a test in https://github.com/rebing/graphql-laravel/pull/313 would trigger a bug if the extra change wouldn't be there.

It's about unwrapping the type when SelectFields traverses _nested_ relation ships and tries to resolve the type for it.

mfn:
> This PR doesn't include a test for it but a test in #313 would trigger a bug if the extra change wouldn't be there.

#313 has since been merged into master and this PR rebased, showing how it correctly fixes the cases 🤞

PR #295, opened 2019-06-01
TBD: Remove release.sh and drop develop branch

mfn:
To be discussed:
- I propose to drop the `develop` branch
- and also the `release.sh` script

Rational:
- Personally I don't have a use for neither of them and I also don't think they're necessary
- Having the `develop` branch just adds a slight management overhead, I don't see a practical purpose it currently serves
- For releases, I just directly create the tags on github from a branch (`master` or `v1`) directly, e.g. https://github.com/rebing/graphql-laravel/releases/tag/v1.22.0 was created that way
- It encourages development on `master` and hopefully also that much more tests will be added in the future to maintain stability (tests are supposed to be green at all times in master).

There might be place for a replacement of `release.sh` in the future, for example automating updating the newly `CHANGELOG.md`, but there are no specific plans at the moment.


mfn:
Also deleted the `develop` branch

PR #302, opened 2019-06-02
composer: remove and ignore composer.lock

mfn:
It's possible to use it for libraries, but not necessary and I believe
it's better to not have it.

Libraries should have the broadest support of dependencies so actual
users of the libraries face less version constraints. This should also
apply to developers of this library.

The composer.lock _only_ applies when developing this library and it's
irrelevant for users of this library.


mfn:
So removing the lock actually found and issue with a follow-up consequence:
- being free of the `.lock` file and the version constraints, the tests for 5.8 failed because the matching orchestra version wasn't in the constrains => fixed
- an immediate follow-up issue was the changed signature of PHPUnits `setUp`. Luckily it seems this change is backwards compatible for our tests

mfn:
Merged via https://github.com/rebing/graphql-laravel/commit/95feeb963c1d4bc3d493ddffb8e7c0bbc38f30cf

mfn:
> Merged via [95feeb9](https://github.com/rebing/graphql-laravel/commit/95feeb963c1d4bc3d493ddffb8e7c0bbc38f30cf)

Unfortunately "not really", I merged a same-named but outdated local branch and didn't realize it => follow up fix via https://github.com/rebing/graphql-laravel/pull/304 . Ops.

PR #305, opened 2019-06-03
[VALIDATIONERROR] Improve handling

mfn:
## Summary
- Add explicit dependency on illuminate/contracts in composer
- rely on contracts on boundaries to define the interface
- removes ambiguity on when getValidatorMessages returns a MessseBag or
  an array => it now always returns a MessageBag
- introduce getter and make internal field private


mfn:
Blocked until https://github.com/rebing/graphql-laravel/pull/304 is merged

PR #309, opened 2019-06-03
Add PHP types to console commands were possible

mfn:
Only touch non-inherited methods for now


src/Rebing/GraphQL/Console/PublishCommand.php
```diff
@@ -49,12 +49,7 @@ public function __construct(Filesystem $files)
         ];
     }
 
-    /**
-     * Execute the console command.
-     *
-     * @return mixed
-     */
```
mfn: removed phpdoc as it didn't add anything useful

PR #319, opened 2019-06-11
Fix args in sub queries

crissi:
passes the test, but this might not work on sub-sub queries


crissi:
I will try to come up with a more complete solution

PR #331, opened 2019-06-16
[BREAKING CHANGE] Add PHP types / phpdoc

mfn:
This PR does its best to adds all the missing types in the library.

This is a breaking change because anyone using this library will have to adapt methods of classes they derived.

This change is done to:
- improve quality of the code
- improve static analysis capabilities
- increase self-documentation of code and make contribution to and use of this library easier

Where technically possible I used native types, otherwise I added them to the phpdoc.

There's only one change which is not a "type change": in `Field::getAttributes` the `isset()` was removed => reason was that after adding the type, phpstan complained the isset is unnecessary (truly so)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -166,7 +166,7 @@ class UserType extends GraphQLType
         'model'         => User::class,
     ];
 
-    public function fields()
+    public function fields(): array
     {
         return [
             'id' => [
--- a/src/Rebing/GraphQL/GraphQL.php
+++ b/src/Rebing/GraphQL/GraphQL.php
@@ -254,7 +254,11 @@ class GraphQL
         $this->mergeSchemas($name, $schema);
     }
 
-    public function mergeSchemas($name, $schema)
+    /**
+     * @param  string  $name
+     * @param  Schema|array  $schema
+     */
+    public function mergeSchemas(string $name, $schema): void
     {
         if (isset($this->schemas[$name]) && $this->schemas[$name]) {
             $this->schemas[$name] = array_merge_recursive($this->schemas[$name], $schema);
--- a/src/Rebing/GraphQL/GraphQL.php
+++ b/src/Rebing/GraphQL/GraphQL.php
@@ -206,7 +206,12 @@ class GraphQL
         return $objectType;
     }
 
-    protected function buildObjectTypeFromClass($type, $opts = [])
+    /**
+     * @param  ObjectType|string  $type
+     * @param  array  $opts
+     * @return ObjectType
+     */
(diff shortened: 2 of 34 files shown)


src/Rebing/GraphQL/Support/Field.php
```diff
@@ -210,10 +217,7 @@ public function getAttributes()
             'args' => $this->args(),
         ], $attributes);
 
-        $type = $this->type();
-        if (isset($type)) {
-            $attributes['type'] = $type;
-        }
+        $attributes['type'] = $this->type();
```
mfn: This is the code change mentioned in the PR description; a side effect of adding the type and phpstan error reporting.

Readme.md
```diff
@@ -387,20 +387,20 @@ class UpdateUserEmailMutation extends Mutation
...
+    public function args(): array
     {
         return [
             'id' => ['name' => 'id', 'type' => Type::string()],
             'email' => ['name' => 'email', 'type' => Type::string()]
         ];
     }
 
-    public function rules(array $args = [])
+    protected function rules(array $args = []): array
```
mfn: Not a change, only a correction

tests/Database/MutationValidationUniqueWithCustomRulesTests/MutationWithCustomRuleWithRuleObject.php
```diff
@@ -13,12 +13,12 @@ class MutationWithCustomRuleWithRuleObject extends Mutation
...
     ];
 
-    public function type()
+    public function type(): Type
     {
         return Type::nonNull(Type::string());
     }
 
-    public function rules(array $args = [])
+    protected function rules(array $args = []): array
```
mfn: There are a few visibility changes on `rules` but they're actually a correction because the parent has been `protected` all the time

src/Rebing/GraphQL/GraphQL.php
```diff
@@ -118,14 +138,14 @@ public function queryAndReturnResult($query, $params = [], $opts = [])
...
-    public function addTypes($types)
+    public function addTypes(array $types): void
     {
         foreach ($types as $name => $type) {
             $this->addType($type, is_numeric($name) ? null : $name);
         }
     }
 
-    public function addType($class, $name = null)
+    public function addType(string $class, string $name = null): void
```
mfn: This is actually a unintended break 💥 

The code within says:
```php
        if (! $name) {
            $type = is_object($class) ? $class : app($class);
            $name = $type->name;
        }
```

So objects are supported for `$class` too.

We've no test for this but my experiments with phpstan caught this => will follow-up with a fix soon

PR #332, opened 2019-06-17
Handle x-csrf-token changes via response headers

kylekatarnls:
This allows to tests queries in Graphiql even if VerifyCsrfToken is enabled with tokens passed via headers.

PR #335, opened 2019-06-18
[GRAPHIQL] Update to 0.13.0

mfn:
This upgrades from 0.10.1 to [0.13.0](https://github.com/graphql/graphiql/releases/tag/v0.13.0)

There's a 0.13.1 release but those release files were not published on the CDN (yet?).

Upgrading is a manual process currently. I've kept the existing modifications:
- configurable fetch path
- CSRF integration (including the [recent PR with the additional header](https://github.com/rebing/graphql-laravel/pull/332))

I've also tried to **not** change the formatting of the HTML part as to make future updates easier to reason.

I've tested this by copypasting the template into an existing project and it worked 👍 

Here's the paste from the README.md:

----
# GraphiQL integration

See https://github.com/graphql/graphiql for the home of GraphiQL

The file `graphiql.php` is an integration of the example provided by GraphiQL. It's a slightly modified version from the official repository so it can be integrated.

The modifications are documented here in detail as to replicate them on future updates:

- copy example from packages/graphiql/example/index.html
- adjust CDN URLs from https://cdnjs.com/libraries/graphiql
- graphQLFetcher: replace the hardcoded path with the library configured path:
  `return fetch('/graphql', {` => `return fetch('<?php echo $graphqlPath; ?>', {`
- csrf integration
  - right before `function graphQLFetcher` add: `var xcsrfToken = null;`
  - in the actual `fetch` call, add this header: `'x-csrf-token': xcsrfToken || '<?php echo csrf_token(); ?>'`
  - in the first `then` after the `fetch` call: `xcsrfToken = response.headers.get('x-csrf-token');`

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ Next release
   - Test suite has been refactored and now features Database (SQLite) tests too
 
 ### Changed
+- Updated GraphiQL to 0.13.0 (\#335)[https://github.com/rebing/graphql-laravel/pull/335]
+  - If you're using CSP, be sure to allow `cdn.jsdelivr.net` and `cdnjs.cloudflare.com`
 - `ValidatorError`: remove setter and make it a constructor arg, add getter and rely on contracts
 - Replace global helper `is_lumen` with static class call `\Rebing\GraphQL\Helpers::isLumen`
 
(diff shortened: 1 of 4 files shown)


mfn:
@kylekatarnls since you recently contributed to the GraphiQL stuff, do you mind giving this a spin?

PR #336, opened 2019-06-18
Fix typo: xCsrfToken

kylekatarnls:


mfn:
Thanks, but it's not necessary => I'll upgrade to GraphiQL 0.13.0 and in so already fixed the type => https://github.com/rebing/graphql-laravel/pull/335

Issue #215, opened 2019-02-23
Primary key type array


yasergh:
i have models define the primary key id array like this 
**protected $primaryKey = array('id_product', 'id_lang');**
so when make relation ship between to model in graphql display error message 
**Array to string conversion** 
so to sole this issue i change class **SelectFields** in line 72 with this code 
```
        if( ! is_null($primaryKey))
        {
            if(is_array($primaryKey)){
                foreach ($primaryKey as $key){
                    $select[] = $parentTable ? ($parentTable . '.' . $key) : $key;
                }
            }else{
                $primaryKey = $parentTable ? ($parentTable . '.' . $primaryKey) : $primaryKey;
                if( ! in_array($primaryKey, $select))
                {
                    $select[] = $primaryKey;
                }
            }
        }
```
to solve issue can you add this code to an original  file

Best regards

rebing:
#216

PR #341, opened 2019-06-18
[SELECTFIELDS] Remove support for composite primary keys

mfn:
## Summary
phpstan reported this:
Call to function is_array() with string will always evaluate to false

After adding types to `getPrimaryKeyFromParentType()`, it's not possible
to return an array here.

All the information I could find on such composite keys is that Laravel
does not support them (e.g. https://github.com/laravel/framework/issues/5355 ).

I also read a bit through the 5.8 sources and everywhere where
`getKeyName()` is called within Laravel you can see that it's expected
to be a string and doesn't seem like it would truly work with an array.

### More information
This feature was added via https://github.com/rebing/graphql-laravel/pull/216 to resolve https://github.com/rebing/graphql-laravel/issues/215 .

But it doens't go into much detail how this is really achieved and we've no tests for this.

@yasergh can you shed some light on your use cases? How to you make your Laravel version work with composite keys when it doesn't seem to be officially supported? Are you using a library like https://github.com/coenjacobs/eloquent-composite-primary-keys ?

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -27,7 +27,6 @@ parameters:
         - '/Parameter #1 \$haystack of function mb_stripos expects string, \(array\)\|string given/'
         - '/Strict comparison using === between null and array will always evaluate to false/'
         - '/Cannot access offset . on array\|Closure/'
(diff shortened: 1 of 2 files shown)


mfn:
I intend to forward with the removal for v2 soon-ish

mfn:
I'll go forward now and merge this.

Note that I'm happy to re-introduce this with a working examples so that tests can be written for this.

PR #343, opened 2019-06-19
Improve phpstan config

szepeviktor:


phpstan.neon
```diff
@@ -1,12 +1,10 @@
...
   paths:
-  - src
-  - tests
+    - %rootDir%/../../../src/
+    - %rootDir%/../../../tests/
   excludes_analyse:
-  - %rootDir%/../../../src/example/*
-  - %rootDir%/../../../src/resources/views/*
-  - %rootDir%/../../../tests/Support/database/*
+    - %rootDir%/../../../tests/Support/database/
```
mfn: I believe I should switch to `%currentWorkingDir%`; `%rootDir%` doesn't work with Larastan without adjusting which seems odd. currentWorkingDir seems more "stable" as the analyser is always meant to be run from the projects root anyway

szepeviktor: Okay!

mfn:
Added the variable change and force-pushed to be in sync with updated master

mfn:
Thanks @szepeviktor !

szepeviktor:
You're welcome.

PR #347, opened 2019-06-19
Replace Phpstan with Larastan, 2nd attempt

mfn:
Previous one https://github.com/rebing/graphql-laravel/pull/345

I figured the solution with the Laravel version constraint is easy: from all the matrix we only run the static analyser on the latest stable versions, so we simply uninstall Larastan everywhere else. 🎉


szepeviktor:
Here are my 2 _cents_

1. require phpstan/phpstan-shim to avoid dependency conflicts
1. install Larastan and PHPStan only in Travis config

szepeviktor:
still grouping and commenting ignores may help a lot in the future

mfn:
> require phpstan/phpstan-shim to avoid dependency conflicts

But the conflicts are with larastan itself.

> install Larastan and PHPStan only in Travis config

I get your point but I decided against it, because the intention is that developers have the tools right out of the box.

The solution is fine because only one of the builds is run via the analyzer

PR #352, opened 2019-06-20
Remove {graphql_schema?} from graphqil.prefix and thus making it compatible with Lumen

mfn:
I installed this under Lumen 5.8 and `/graphql` endpoint itself worked but `/graphiql` didn't.

After some forth and back I figured it's this part in the `graphql.graphiql.prefix`. Once I removed it, graphiql worked; also with explicit schema names, e.g. `/graphiql/default`.

This seems to be fine (ignored?) in Laravel (also tried to remove it in a Laravel installation and it broken nothing).

Anyone using Lumen can test/verify this?

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -167,7 +167,7 @@ return [
      * Config for GraphiQL (see (https://github.com/graphql/graphiql).
      */
     'graphiql' => [
-        'prefix'     => '/graphiql/{graphql_schema?}',
+        'prefix'     => '/graphiql',
         'controller' => \Rebing\GraphQL\GraphQLController::class.'@graphiql',
         'middleware' => [],
         'view'       => 'graphql::graphiql',


hoatructhanhlam:
I'm having this error with lumen 5.8 
```
Array to string conversion in /Users/dudu/code/zeko_api/vendor/rebing/graphql-laravel/src/Rebing/GraphQL/GraphQLController.php:20
```
This is result of log `$request->request->all()`
```
array (
  'operationName' => NULL,
  'variables' => 
  array (
    'longitude' => 105.8239852,
    'latitude' => 20.9847755,
    'province' => 'Hà Nội',
    'type' => 0,
    'page' => 1,
  ),
  'query' => 'query ($longitude: Float, $latitude: Float, $province: String, $type: Int, $page: Int) {
  restaurants(longitude: $longitude, latitude: $latitude, province: $province, type: $type, page: $page) {
    data {
      id
      name
      avatar {
        thumb
        origin
        __typename
      }
      address
      opening_times {
        start
        end
        __typename
      }
      location {
        coordinates
        __typename
      }
      province
      district
      price_range {
        min
        max
        __typename
      }
      distance
      __typename
    }
    current_page
    __typename
  }
}
',
)  
```

As we see, request is nested array, so `implode` function cannot working

mfn:
@hoatructhanhlam this seems unrelated to this PR:
- please check if your issue is present => https://github.com/rebing/graphql-laravel/labels/Lumen
- if not, please file a new issue using the template
  (if you do so, please also add the full HTTP request URL you used)

mfn:
@hoatructhanhlam but it would be interesting to know if GraphiQL works for you for Lumen with/without this change

hoatructhanhlam:
> @hoatructhanhlam but it would be interesting to know if GraphiQL works for you for Lumen with/without this change

Oh, graphiql worked well with this change for lumen.
It is true that my problem is not related to this pull request. I will create a other pull request to resolve this problem

mfn:
@hoatructhanhlam 

> Oh, graphiql worked well with this change for lumen.

Thank you.

To be very explicit: without this change, GraphiQL **did not** work for you, is this right ;-)

Just to be double-sure it's not my testing environment 😄

mfn:
I'm not so sure anymore this was a good idea to remove it.

I believe the underlying problem was with https://github.com/rebing/graphql-laravel/pull/358 and that this should be revisited.

As of right now, custom schemas are not passed through but should :|

PR #353, opened 2019-06-21
Use sql expression in alias

crissi:
missing some docs, will add if the feature get accepted


mfn:
Have you tried directly using `DB::raw()` inside the alias?

See:
- https://github.com/rebing/graphql-laravel/issues/280#issuecomment-496156750
- https://github.com/rebing/graphql-laravel/issues/272

crissi:
yes

crissi:
> Have you tried directly using `DB::raw()` inside the alias?
> 
> See:
> 
> * [#280 (comment)](https://github.com/rebing/graphql-laravel/issues/280#issuecomment-496156750)
> * #272

yes

src/Support/SelectFields.php
```diff
@@ -294,6 +295,26 @@ private static function isQueryable(array $fieldObject): bool
...
+     */
+    private static function getColumn(string $key, $fieldObject)
+    {
+        if (isset($fieldObject->config['alias'])) {
+            $alias = $fieldObject->config['alias'];
+
+            if (is_string($alias)) {
+                return $alias.' AS '.$key;
+            } elseif (is_callable($alias)) {
+                return DB::raw('('.$alias().') AS '.$key);
```
mfn: I may be wrong but generating a fragment like `DB::raw('('.$alias().') AS '.$key)` will not properly fully quality `$key` and there are databases like PostgreSQL which would need wrapping in `"` if the name contains upper case characters. This is just one case which comes to my mind.

crissi: your solution seems more straith forward, even though it is a bit more work, so closing this

PR #354, opened 2019-06-21
Add actual Laravel/Lumen integration with a basic test (and fix Lumen)

mfn:
## Summary
I wanted to have a and very small test whether the integration into Laravel/Lumen works on the most basic level. After I found out about how [Larastan does it](https://github.com/nunomaduro/larastan/blob/master/tests/laravel-test.sh), I decided to give it a go.

Similar to static analysis which is performed for a single matrix combination (PHP 7.3, Laravel 5.8), for this entry we now perform an actual integration test for Laravel and Lumen.

Both work the same way (but need different approaches):
- Install Framework
- Install graphql-laravel from the commit under test in Travis
- configure installation
- test basic query execution works
- test GraphiQL loads

Along the way I found out that Lumen in `master` was broken due to the [recent addition of type declarations](https://github.com/rebing/graphql-laravel/pull/331), therefore some small code adaptions were also necessary.

I also learned that GraphiQL didn't work for Lumen because it calls `csfr_token()` function which isn't available to Lumen by default => a small entry into the README Lumen installation part has been added (and the integration test also performs this workaround steps).

### DIFF
--- a/tests/integration-larvavel.sh
+++ b/tests/integration-larvavel.sh
@@ -1,0 +1,56 @@
+#!/usr/bin/env bash
+
+# Inspired by https://github.com/nunomaduro/larastan/blob/669b489e10558bd45fafc2429068fd4a73843802/tests/laravel-test.sh
+#
+# Create a fresh Laravel installation, install our package in it and run some
+# basic tests to ensure everything works.
+#
+# This script is meant to be run on Travis CI
+
+echo "Prevent shallow repository error"
+git fetch --unshallow
+
+echo "Install Laravel"
+travis_retry composer create-project --quiet --prefer-dist "laravel/laravel" laravel
+cd laravel
+
+echo "Add package from source"
+sed -e 's|"type": "project",|&\n"repositories": [ { "type": "vcs", "url": "../" } ],|' -i composer.json
+travis_retry composer require --dev "rebing/graphql-laravel:dev-master#${TRAVIS_COMMIT}"
+
+echo "Publish vendor files"
+php artisan vendor:publish --provider="Rebing\GraphQL\GraphQLServiceProvider"
+
+echo "Make GraphQL ExampleQuery"
+php artisan make:graphql:query ExampleQuery
+
+echo "Add ExampleQuery to config"
+sed -e "s|// 'example_query' => ExampleQuery::class,|\\\App\\\GraphQL\\\Query\\\ExampleQuery::class,|" -i config/graphql.php
+
+echo "Start Webserver"
+php -S 127.0.0.1:8001 -t public >/dev/null 2>&1 &
+sleep 2
+
+echo "Send GraphQL HTTP request to fetch ExampleQuery"
+curl 'http://127.0.0.1:8001/graphql?query=%7BExampleQuery%7D' -sSfLv | grep 'The ExampleQuery works'
+
+if [[ $? = 0 ]]; then
+  echo "Example GraphQL query works 👍"
+else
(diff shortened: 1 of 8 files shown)


tests/integration-larvavel.sh
```diff
@@ -0,0 +1,56 @@
...
+
+# Inspired by https://github.com/nunomaduro/larastan/blob/669b489e10558bd45fafc2429068fd4a73843802/tests/laravel-test.sh
+#
+# Create a fresh Laravel installation, install our package in it and run some
+# basic tests to ensure everything works.
+#
+# This script is meant to be run on Travis CI
+
+echo "Prevent shallow repository error"
+git fetch --unshallow
```
mfn: This refers to this https://travis-ci.org/rebing/graphql-laravel/jobs/548915299#L621-L637

One has to see it to understand it, I guess. It only happened for the "PR" build, not for branch builds.

tests/integration-larvavel.sh
```diff
@@ -0,0 +1,56 @@
...
+
+echo "Make GraphQL ExampleQuery"
+php artisan make:graphql:query ExampleQuery
+
+echo "Add ExampleQuery to config"
+sed -e "s|// 'example_query' => ExampleQuery::class,|\\\App\\\GraphQL\\\Query\\\ExampleQuery::class,|" -i config/graphql.php
+
+echo "Start Webserver"
+php -S 127.0.0.1:8001 -t public >/dev/null 2>&1 &
+sleep 2
```
mfn: Not sure, could be `sleep 1`. I couldn't find a ready-available shell helper to wait until the web server is available but since startup time should be way <1s I figured let' make it simple.

tests/integration-larvavel.sh
```diff
@@ -0,0 +1,56 @@
...
+php artisan vendor:publish --provider="Rebing\GraphQL\GraphQLServiceProvider"
+
+echo "Make GraphQL ExampleQuery"
+php artisan make:graphql:query ExampleQuery
+
+echo "Add ExampleQuery to config"
+sed -e "s|// 'example_query' => ExampleQuery::class,|\\\App\\\GraphQL\\\Query\\\ExampleQuery::class,|" -i config/graphql.php
+
+echo "Start Webserver"
+php -S 127.0.0.1:8001 -t public >/dev/null 2>&1 &
```
mfn: I had some issues getting a proper output from curl afterwards so I silenced this completely. Might revisit this in the future.

tests/integration-larvavel.sh
```diff
@@ -0,0 +1,56 @@
...
+sleep 2
+
+echo "Send GraphQL HTTP request to fetch ExampleQuery"
+curl 'http://127.0.0.1:8001/graphql?query=%7BExampleQuery%7D' -sSfLv | grep 'The ExampleQuery works'
```
mfn: The flags `sSfLv` are a random internet sample to get enough output to hopefully easier understand failures.

tests/integration-larvavel.sh
```diff
@@ -0,0 +1,56 @@
...
+
+if [[ $? = 0 ]]; then
+  echo "Example GraphQL query works 👍"
+else
+  echo "Example GraphQL query DID NOT work 👎"
+  curl 'http://127.0.0.1:8001/graphql?query=%7BExampleQuery%7D' -sSfLv
```
mfn: Simple send it again, this time without `grep` so the full output is "visible" in the build logs.

Same for the `cat storage/logs/*` below.

Issue #236, opened 2019-03-22
compatibility with lumen 5.8


MarceloPetrucio:
Hello everyone,

I'm having internal problems in this package, I don't know exactly where the problem is because it is not returning the error of where it is going wrong.

I did all the installation according to the initial readme and had no problem installing, but when I try to do a query, or a post on the route "localhost:4000/graphql" or try to run the GraphQL - Playground it doesn't recognize the server

Has anyone gone through this or know where I can look to find out the problem?

rebing:
Please specify the error

pxgamer:
Is there anything in the `storage/logs`? I've managed to get it working on Lumen 5.8, except `Introspection` queries result in a 500, so not sure if this is what you're referring to?

renepardon:
HI @rebing 
I tried this package today for the first time with lumen. 

**My installation:**

- Lumen (5.8.4) (Laravel Components 5.8.*)
- thedevsaddam/lumen-route-list
- predis/predis

I copied the example files and adjusted namespaces. Also added them to corresponding place within the graphql.php config file. But even with commented configuration schema it doesn't work for unknown reason. Maybe you can have a look and see what's going wrong here. :)

**The error:**

```
[2019-04-23 12:51:34] local.ERROR: Rebing\GraphQL\Exception\SchemaNotFound: Type 
    query IntrospectionQuery {
      __schema {
        queryType { name }
        mutationType { name }
        subscriptionType { name }
        types {
          ...FullType
        }
        directives {
          name
          description
          locations
          args {
            ...InputValue
          }
        }
      }
    }
    fragment FullType on __Type {
      kind
      name
      description
      fields(includeDeprecated: true) {
        name
        description
        args {
          ...InputValue
        }
        type {
          ...TypeRef
        }
        isDeprecated
        deprecationReason
      }
      inputFields {
        ...InputValue
      }
      interfaces {
        ...TypeRef
      }
      enumValues(includeDeprecated: true) {
        name
        description
        isDeprecated
        deprecationReason
      }
      possibleTypes {
        ...TypeRef
      }
    }
    fragment InputValue on __InputValue {
      name
      description
      type { ...TypeRef }
      defaultValue
    }
    fragment TypeRef on __Type {
      kind
      name
      ofType {
        kind
        name
        ofType {
          kind
          name
          ofType {
            kind
            name
            ofType {
              kind
              name
              ofType {
                kind
                name
                ofType {
                  kind
                  name
                  ofType {
[...]

Issue #204, opened 2019-02-18
Multiple schema is not working in lumen 


sh4msi:
hi

req url = http://localhost:8000/api/secret

```
mutation {
  createUser(
    name: "tester"
    email:"123@test.com"
    password:"123"
  ) {
    id,
    email
  }
}
```

error:
`"message": "Schema is not configured for mutations.",`

config:
```
    'schemas' => [
        'default' => [
            'query' => [
                 'login'    => UserLoginQuery::class,
            ],
             'mutation' => [
                //..
            ],
            'method' => ['post'],
        ],
        'secret' => [
            'query' => [
                //..
            ],
            'mutation' => [
                'createUser' => CreateUserMutation::class
            ],
           //  'middleware' => ['auth:api'],
            'method' => ['post'],
        ],
    ],
```

sh4msi:
Hi guys

**[my solution for this problem
](https://github.com/sh4msi/graphql-laravel/commit/bb426dd18bbf51f7b463da7cb78f8ff18c623fa0)**



> Of course, you can, not change the original files and just call the other controller from the config file.

MrGekko:
The controller changed a little bit about 6 days ago. If this is a solution that everyone needs maybe you could create a pull-request?

mfn:
Looks like this is the same as https://github.com/rebing/graphql-laravel/issues/236

mfn:
I've just merged a PR which I believe fixes this problem.

There's no release yet so if you're eager, you need to test with `dev-master`.
Also be aware that the master branch has other changes we may require further adaptions to make it work (types mostly).

If it still doens't work, please create a new issue with details to reproduce it; thank you!

PR #358, opened 2019-06-22
[LUMEN] Fix path multi-level support

mfn:
## Summary
The existing code didn't work when accessing `/graphql/some/schema`,
only for `/graphql/some`.

With the current routing setup, the parameters to the route describe
the schema to use.

The existing code did check `request->count()` which is something else,
it contains e.g. query or JSON parameters and didn't return a correct
schema.

Route parameters are actually available in `$route[2]` which the code
now checks for.

This also means GraphiQL now finally fully works. The reason is that
GraphiQL sent multiple json parameters (operationName, query, variables)
which the code due to `request->all()` used to build the schema name
which never worked correctly.

It's not possible to write a test for this as orchestra/testbench
does not support Lumen currently.

Interestingly https://github.com/rebing/graphql-laravel/issues/236#issuecomment-486116951 was almost the solution though not entirely (the `->count()` check was still not correct).

### Links
- Fixes https://github.com/rebing/graphql-laravel/issues/204
- Fixes https://github.com/rebing/graphql-laravel/issues/236

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ CHANGELOG
 - Replace global helper `is_lumen` with static class call `\Rebing\GraphQL\Helpers::isLumen`
 
 ### Fixed
+- Path multi-level support for Schemas works again [\#358](https://github.com/rebing/graphql-laravel/pull/358)
 - SelectFields correctly passes field arguments to the custom query [\#327](https://github.com/rebing/graphql-laravel/pull/327)
   - This also applies to privacy checks on fields, the callback now receives the field arguments too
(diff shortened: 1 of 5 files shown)


mfn:
Also added an actual integration test where a path multi level schema is used to execute a query, showing that this works (it's not possible to only test with Lumen in our testsuite, orchestra doesn't support it).

adamlc:
Has this made it into v2? I'm having the same problem in v2 :(

mfn:
This PR, yes.

If it didn't fix it, we still have a bug: please open a new issue with all the details!

PR #360, opened 2019-06-24
[V1][ROUTES] Prefix named graphiql routes with `graphql`

mfn:
Compatibility with Folklore

Technically a breaking change but [named routes were only recently added](https://github.com/rebing/graphql-laravel/pull/264) and I don't think anyone jumped already onto them.


mfn:
[Merged change also in master](https://github.com/rebing/graphql-laravel/commit/6d94ea7a8ba301900e672247657d21da4a1134c3)

PR #374, opened 2019-07-04
Introduce TypeConvertible interface and supporting code

mfn:
- Clarifies use of the defined types and necessary for `Scalar` types
- Increases compatibility with Folklore lib (should be a simple namespace replace and done)


tests/Support/Types/MyCustomScalarString.php
```diff
@@ -28,7 +30,7 @@ public function serialize($value)
     }
 
     /**
-     * Parses an externally provided value (query variable) to use as an input
+     * Parses an externally provided value (query variable) to use as an input.
```
mfn: StyleCI 🤷‍♀️

Issue #377, opened 2019-07-04
Computed properties issue


ZaidBarghouthi:
Consider the following: 

```gql
type User {
   id
   name
}
type Post {
   id
   title
   body
   isPublished
   publishedAt
}
```
Where `isPublished` is defined as a computed property as

```php
class Post extends Model
{
    ...
    public function getIsPublishedAttribute(): bool
    {
        $publishedAt = $this->published_at;
        return $publishedAt !== null;
    }
}
```
where ~~is_published~~ `published_at` is a `dateTime` field in the database.      **Edited**

So, `isPublished` is *not selectable*. The problem is the value of `isPublished` depends on
`published_at` and the following query will **always** give `isPublished : false`

```gql
{
  users {
    id
    name
    posts {
      id
      isPublished #always false
    }
  }
}
```

and 
```gql
{
  users {
    id
    name
    posts {
      id
      publishedAt
      isPublished #correct result
    }
  }
}
```

I suppose we need something like
```php

            'isPublished' => [
                'type' => Type::nonNull(Type::boolean()),
                'selectable' => false,
                'must_select' => 'publishedAt'
            ],
```
because one wouldn't want to select `published_at` always, yet want it to always be selected when `isPublished` is selected

mfn:
Philosophical note: I'm wondering if it's smart to add more and more custom attributes to such a config, mixed with the ones from GraphQL. I thinking about something funneling them in a single attribute, e.g.:
```php
            'isPublished' => [
                'type' => Type::nonNull(Type::boolean()),
                'select_fields' => [
                    'selectable' => false,
                    'must_select' => 'publishedAt',
                ],  
            ],
```
Just thinking out loud and, sorry, unrelated to the issue itself: but I acknowledge a new directive is probably required.

Suggestions:
- naming: `requires`?
- it should probably be an array because who knows how many other columns it requires
- (and it should also support selecting relations because that's the next thing coming up)
- …

mfn:
After one more thought, I'm not sure there is a problem, please see https://github.com/rebing/graphql-laravel/pull/378#issuecomment-508571375

ZaidBarghouthi:
I know always fixes the issue, but its a workaround, 

> because one wouldn't want to select published_at always, yet want it to always be selected when isPublished is selected

ZaidBarghouthi:
That the why i came across #369. I have a lot if computed properties in my project. some of them are morph relations :P

mfn:
[...]

PR #378, opened 2019-07-04
[TESTS] Computed properties evaluation issues

ZaidBarghouthi:
see #377

### DIFF
--- a/tests/Database/SelectFields/ComputedPropertiesTests/ComputedPropertiesTest.php
+++ b/tests/Database/SelectFields/ComputedPropertiesTests/ComputedPropertiesTest.php
@@ -13,6 +13,11 @@ class ComputedPropertiesTest extends TestCaseDatabase
 {
     use SqlAssertionTrait;
 
+    /**
+     * Once https://github.com/rebing/graphql-laravel/issues/377 is fixed,
+     * the test needs to be changed, showing that the computed properties
+     * evaluate correctly, or perhaps throw an error if not properly configured.
+     */
     public function testComputedProperty(): void
     {
         $user = factory(User::class)->create([
@@ -42,11 +47,11 @@ GRAQPHQL;
 
         $result = $this->graphql($query);
 
-        // $this->assertSqlQueries(<<<'SQL'
-        // select "users"."id", "users"."name" from "users";
-        // select "posts"."id", "posts"."user_id" from "posts" where "posts"."user_id" in (?) order by "posts"."id" asc;
-        // SQL
-        // );
+        $this->assertSqlQueries(<<<'SQL'
+select "users"."id", "users"."name" from "users";
+select "posts"."id", "posts"."user_id" from "posts" where "posts"."user_id" in (?) order by "posts"."id" asc;
+SQL
+        );
 
         $expectedResult = [
             'data' => [
--- a/tests/Database/SelectFields/ComputedPropertiesTests/ComputedPropertiesTest.php
+++ b/tests/Database/SelectFields/ComputedPropertiesTests/ComputedPropertiesTest.php
@@ -1,0 +1,90 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rebing\GraphQL\Tests\Database\SelectFields\ComputedPropertiesTests;
+
+use Rebing\GraphQL\Tests\TestCaseDatabase;
+use Rebing\GraphQL\Tests\Support\Models\Post;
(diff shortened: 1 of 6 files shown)


tests/Database/SelectFields/ComputedPropertiesTests/ComputedPropertiesTest.php
```diff
@@ -0,0 +1,90 @@
...
+use Rebing\GraphQL\Tests\TestCaseDatabase;
+use Rebing\GraphQL\Tests\Support\Models\Post;
+use Rebing\GraphQL\Tests\Support\Models\User;
+use Rebing\GraphQL\Tests\Support\Traits\SqlAssertionTrait;
+
+class ComputedPropertiesTest extends TestCaseDatabase
+{
+    use SqlAssertionTrait;
+
+    public function testComputedProperty(): void
```
mfn: Can you please add a phpdoc linking to the issue, similar to what I/we did in `\Rebing\GraphQL\Tests\Database\SelectFields\AlwaysRelationTests\AlwaysRelationTest::testAlwaysSingleHasManyRelationField` et al?

ZaidBarghouthi: sure

ZaidBarghouthi:
@mfn whats a proper name of the test anyway?

tests/Database/SelectFields/ComputedPropertiesTests/UsersQuery.php
```diff
@@ -0,0 +1,37 @@
...
+        return Type::listOf(GraphQL::type('User'));
+    }
+
+    public function resolve($root, $args, $contxt, ResolveInfo $info, Closure $getSelectFields)
+    {
+        /** @var SelectFields $selectFields */
+        $selectFields = $getSelectFields();
+
+        return User
+            ::query()
```
mfn: I think this can just start with `::select(…)`, not requiring the `::query()` at all

tests/Support/Models/Post.php
```diff
@@ -15,11 +16,17 @@
  * @property string|null $body
  * @property int|null $user_id
  * @property bool $flag
+ * @property Illuminate\Support\Carbon|null $published_at
+ * @property bool $is_published
```
mfn: Nice 👍 

I did evaluate to bring Laravels ide-helper packer on board here but I think the outcome was that it doesn't support packages (or would need too much work to be made work or something).

tests/Database/SelectFields/ComputedPropertiesTests/ComputedPropertiesTest.php
```diff
@@ -0,0 +1,90 @@
...
+
+        $this->sqlCounterReset();
+
+        $result = $this->graphql($query);
+
+//         $this->assertSqlQueries(<<<'SQL'
+// select "users"."id", "users"."name" from "users";
+// select "posts"."id", "posts"."user_id" from "posts" where "posts"."user_id" in (?) order by "posts"."id" asc;
+// SQL
+//         );
```
mfn: Btw, don't see a reason to not simple enable them; it's fine if they are not showing what is _really_ expected.

For now they're a snapshot of the current situation.

mfn:
Did you try adding `'always' => 'published_at',` to the `'isPublished'` configuration?

This seems to work for me.

tests/Database/SelectFields/ComputedPropertiesTests/ComputedPropertiesTest.php
ZaidBarghouthi: yes, that was an oversight

ZaidBarghouthi:
> Did you try adding `'always' => 'published_at',` to the `'isPublished'` configuration?
> 
> This seems to work for me.

Yea that works. The issue in my opinion is it resolves it incorrectly if always in not set. Its a bit of a hidden issue.

tests/Database/SelectFields/ComputedPropertiesTests/UsersQuery.php
ZaidBarghouthi: 👍

mfn:
> Yea that works. The issue in my opinion is it resolves it incorrectly if always in not set. Its a bit of a hidden issue.

Yeah, sorry, you're right. You even described. I guess my mind wandered off somewhere.

PR #389, opened 2019-07-06
Remove Fluent dependency on core `Type`

mfn:
This is an experiment, because it seems the dependency on Fluent doesn't
bring much to the table.

At least according to the test suite, the only missing piece was a
property __get shim.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ CHANGELOG
   - Previously the initial query arguments would be used everywhere
 
 ### Removed
+- Removed `\Fluent` dependency on `\Rebing\GraphQL\Support\Type`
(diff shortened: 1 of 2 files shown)


mfn:
Benchmark time!

Recently (unrelated to this PR) performance issues were reported: https://github.com/rebing/graphql-laravel/issues/390

With a repro with 5k queries and 5k types: https://github.com/dnyg/graphql-laravel-performancetest

Output of the following:
```php
        dump(memory_get_peak_usage());
        dump(memory_get_usage());
```

With Fluent:
```
94_460_400
94_274_112
```

_Without_ Fluent:
```
91_100_088
90_913_800
```

**Conclusion:** without Fluent, roughly 3.5% **less** peak memory usage

Issue #390, opened 2019-07-06
Performance scaling issues


dnyg:
In my project I have about 200 mutations, 200 queries and 400 types, and it looks like request performance degrades for every type you add.

I've created a repo to demonstrate: https://github.com/dnyg/graphql-laravel-performancetest

It contains a clean Laravel 5.8 installation with graphql-laravel installed. It has 5000 types and 5000 queries which all returns a simple hardcoded result (no database required for testing):

```
public function resolve($root, $args, $context, ResolveInfo $resolveInfo)
{
    return [
        'id' => 1
    ];
}
```

Obviously the retrospection call takes a long time, but the requests themselves are also painfully slow. The simple query below returns a hardcoded value, but it takes on average 1.93s to execute on my local machine.

```
{
  test {
    id
  }
}
```

If I remove the other queries and types, it executes in 162ms on average.

What can be done to resolve this issue to the application can scale while maintaining reasonable performance?

_The Xdebug profiler result hinted that the request spent 38% of the request time running the function `buildObjectTypeFromClass` in Rebing\GraphQL\GraphQL.php` in the original project - I imagine the number is much higher in the test-repo._

mfn:
Repos are always awesome for reproducing 👍 

Two things upfront:
- I tested with `mater`, not v1 releases, which have some divergence (but nothing related to performance)
  - therefore some type adjustments needed to be made
- I've added a command to run said query as to have something easily to test from command lin

```php
    public function handle()
    {
        $graphql = <<<'GRAPHQL'
{
  test {
    id
  }
}
GRAPHQL;

        $result = \Rebing\GraphQL\Support\Facades\GraphQL::query($graphql);

        echo json_encode($result, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES), "\n"; exit;
    }
```
I've then ran this using `PHP 7.3.5 (cli)`:
` /usr/local/bin/php -dzend_extension=/usr/local/lib/php/pecl/20180731/xdebug.so -d xdebug.profiler_enable=1 -d xdebug.profiler_output_dir=./  artisan graphql:query`

The cachegrind output is ~76MB uncompressed => [cachegrind.out.29514.zip](https://github.com/rebing/graphql-laravel/files/3365055/cachegrind.out.29514.zip)

In this scenario `buildObjectTypeFromFields` pops up, but it's not the only one: `GraphQL\Type\Schema::__construct()` also takes significant time:
[...]

PR #394, opened 2019-07-08
lazy load types

crissi:
This only works if types are added as "key" => Type::class in config, since bootTypes::bootTypes will instanciate a new entry of all types if it can not associate the name directly. See addType in GraphQL.php

### Links
- Related to https://github.com/rebing/graphql-laravel/issues/390

### DIFF
--- a/src/GraphQL.php
+++ b/src/GraphQL.php
@@ -22,6 +22,8 @@ use Rebing\GraphQL\Exception\SchemaNotFound;
 use Illuminate\Contracts\Container\Container;
 use Illuminate\Contracts\Debug\ExceptionHandler;
 use Rebing\GraphQL\Support\Contracts\TypeConvertible;
+use Rebing\GraphQL\Support\UploadType;
+use GraphQL\Type\Introspection;
 
 class GraphQL
 {
@@ -35,6 +37,9 @@ class GraphQL
      * @var array<string,object|string>
      */
     protected $types = [];
+
+    protected const UPLOAD_TYPE_NAME = 'Upload';
+
     /** @var Type[] */
     protected $typesInstances = [];
 
@@ -60,7 +65,6 @@ class GraphQL
         $schemaQuery = Arr::get($schema, 'query', []);
         $schemaMutation = Arr::get($schema, 'mutation', []);
         $schemaSubscription = Arr::get($schema, 'subscription', []);
-        $schemaTypes = Arr::get($schema, 'types', []);
 
         $query = $this->objectType($schemaQuery, [
             'name' => 'Query',
@@ -75,11 +79,36 @@ class GraphQL
         ]);
 
         return new Schema([
-            'query'         => $query,
-            'mutation'      => ! empty($schemaMutation) ? $mutation : null,
-            'subscription'  => ! empty($schemaSubscription) ? $subscription : null,
-            'typeLoader' => function($name) {
-                return $this->type($name);
+            'query' => $query,
+            'mutation' => !empty($schemaMutation) ? $mutation : null,
+            'subscription' => !empty($schemaSubscription) ? $subscription : null,
+            'typeLoader' => function ($name) {
+                if ($name === self::UPLOAD_TYPE_NAME) {
+                    return UploadType::getInstance();
+                }
+                $keyName = array_search($name, $this->typesInstances);
+
+                return $this->type($keyName);
+            },
+            // The closure is only called when loading the whole schema
+            'types' => function () use ($schema) {
+                $types = [];
+                $schemaTypes = Arr::get($schema, 'types', []);
+
+                if (count($schemaTypes)) {
+                    foreach ($schemaTypes as $name => $type) {
+                        $objectType = $this->objectType($type, is_numeric($name) ? [] : [
+                            'name' => $name,
+                        ]);
+                        $this->typesInstances[$name] = $objectType;
+                        $types[] = $objectType;
+                    }
+                } else {
+                    foreach ($this->getTypes() as $name => $type) {
+                        $types[] = $this->type($name);
+                    }
+                }
+                return $types;
             }
         ]);
     }
--- a/src/GraphQL.php
+++ b/src/GraphQL.php
@@ -22,7 +22,6 @@ use Rebing\GraphQL\Exception\SchemaNotFound;
 use Illuminate\Contracts\Container\Container;
 use Illuminate\Contracts\Debug\ExceptionHandler;
 use Rebing\GraphQL\Support\Contracts\TypeConvertible;
-use Rebing\GraphQL\Support\UploadType;
 
 class GraphQL
 {
@@ -77,13 +76,14 @@ class GraphQL
 
         return new Schema([
             'query' => $query,
-            'mutation' => !empty($schemaMutation) ? $mutation : null,
-            'subscription' => !empty($schemaSubscription) ? $subscription : null,
+            'mutation' => ! empty($schemaMutation) ? $mutation : null,
+            'subscription' => ! empty($schemaSubscription) ? $subscription : null,
             'typeLoader' => function ($name) {
                 $keyName = array_search($name, $this->typesInstances, true);
                 if ($keyName !== false) {
                     return $this->type($keyName);
                 }
+
                 return $this->type($name);
             },
             // The closure is only called when loading the whole schema
@@ -104,8 +104,9 @@ class GraphQL
                         $types[] = $this->type($name);
                     }
                 }
+
                 return $types;
-            }
+            },
         ]);
     }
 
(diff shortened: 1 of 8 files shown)


src/GraphQL.php
```diff
@@ -55,33 +55,13 @@ public function schema($schema = null): Schema
 
         $this->clearTypeInstances();
 
-        foreach ($this->getTypes() as $name => $type) {
-            $this->type($name);
-        }
-
```
mfn: I forgot to mention that yesterday I also removed this code because I also had the feeling it wasn't doing much 😄

crissi: Tested on a live system and failed.

The type loader receives the name from the Types name-attribute wich can be different from the alias used in config. That mean we have to check every file for the name.

One way, could be to make a artisan graphql cache command that mapped the type names to the classes so it was easily found.

But I might be missing something...

mfn: > One way, could be to make a artisan graphql cache command that mapped the type names to the classes so it was easily found.

This direction seem appropriate to explore!

Basically we need to figure out what and how can we cache it and then make a config entry for this.

With such a thing in place, if you deploy code (changes to GraphQL type system in some way) and don't clear the cache it would probably break. Which is _fine_ because the recommendation _is_ to clear the cache after deployment.

src/GraphQL.php
```diff
@@ -94,11 +73,37 @@ public function schema($schema = null): Schema
...
         return new Schema([
-            'query'         => $query,
-            'mutation'      => ! empty($schemaMutation) ? $mutation : null,
-            'subscription'  => ! empty($schemaSubscription) ? $subscription : null,
-            'types'         => $types,
+            'query' => $query,
+            'mutation' => !empty($schemaMutation) ? $mutation : null,
+            'subscription' => !empty($schemaSubscription) ? $subscription : null,
+            'typeLoader' => function ($name) {
+                $keyName = array_search($name, $this->typesInstances);
```
mfn: We should try to avoid `array_search` as it has linear performance characteristics.

mfn:
I've rebased with master (fixed styleci, adapted phpstan) where I just added test for file uploads.

Really just a coincidence, but these tests now break :|

(guess phpstan was right after all about the return value from `array_search` and I should not have put it into `ignoreErrors`?)

crissi:
> I've rebased with master (fixed styleci, adapted phpstan) where I just added test for file uploads.
> 
> Really just a coincidence, but these tests now break :|
> 
> (guess phpstan was right after all about the return value from `array_search` and I should not have put it into `ignoreErrors`?)

so far I don't know how to fix the upload type (custom scalar types), the name I get in is Upload, but I have know Idea where to look for the class name

crissi:
> Do we have any recent performance numbers (is it working 😄 )?

nope, might get to benchmark a bit tomorrow

mfn:
Rebased with master, applied StyleCI fixes and adapted phpstan.

Btw: StyleCI violations are easy to fix => click on the link => download the raw diff => apply patch (I use PhpStorm which has a feaure "apply patch from clipboard")

src/Support/UploadType.php
```diff
@@ -61,4 +63,9 @@ public static function getInstance()
 
         return $inst;
     }
+
+    public function toType(): Type
+    {
+        return self::getInstance();
```
mfn: I suggest ~self~ `static`

src/GraphQLServiceProvider.php
```diff
@@ -77,6 +78,8 @@ protected function bootPublishes(): void
     protected function bootTypes(): void
     {
         $configTypes = config('graphql.types');
+
+        $configTypes['Upload'] = UploadType::class;
```
mfn: Ah, I see!

Now instead of using this:
```php
    public function args(): array
    {
        return [
            'files' => [
                'type' => Type::listOf(UploadType::getInstance()),
            ],
        ];
    }
```
we can use
```php
                'type' => Type::listOf(GraphQL::type('Upload')),
```
which also is more idiomatic how things are defined.

However my feeling is we should not add this hardcoded: not everyone might want to use the upload capability.

In fact, it's the same with the upload middleware (always enabled).

I feel there needs to be some refactoring for the middleware _and_ the `Upload` type should be added by the user manually if they want to use it (effectively making this a documentation issue).

FTR I'm also toying with the idea removing the whole upload middleware from this package in its current form and rather use https://github.com/Ecodev/graphql-upload and have an optionally enabled adapter/bridge.

mfn: > I feel there needs to be some refactoring for the middleware and the Upload type should be added by the user manually if they want to use it (effectively making this a documentation issue).

FTR I'm talking about _outside_ this PR!

mfn: For the UploadType change, see https://github.com/rebing/graphql-laravel/pull/401

src/GraphQL.php
crissi: > We should try to avoid `array_search` as it has linear performance characteristics.

waiting to change this, to after I have done some benchmarking

src/GraphQLServiceProvider.php
mfn: The UploadType change was merged, thus I've rebased this PR _and_ removed the hardcoded UploadType, as it isn't necessary anymore.

src/Support/UploadType.php
mfn: Obsolete, handled in master now (and differently)

crissi:
thought I fixed it with the names being diffent in class queries, must investigate again

mfn:
Can you elaborate a bit on the last commit/change?

I'm just wondering why no test broke before that change (was under the impression these sort of things were covered).

crissi:
No, there was no types in the tests covering types that had a different config name than it equalient class type name

Få Outlook til iOS<https://aka.ms/o0ukef>
Sendt: Thursday, July 11, 2019 7:59:33 PM
Til: rebing/graphql-laravel
Cc: Christian; Author
Emne: Re: [rebing/graphql-laravel] [WIP] lazy load types (#394)


Can you elaborate a bit on the last commit/change?

I'm just wondering why no test broke before that change (was under the impression these sort of things were covered).

crissi:
So right now it is alot faster if the config alias and the class name alias matches, but we have no efficient way to check that in runtime. So we can not rely on that!

So the only step  I can come up with is to make a cache artisan command. I might get to work on that in the weekend.

Do you have any other suggestions or comments?

src/GraphQL.php
mfn: Do you have any results?

mfn:
> No, there was no types in the tests covering types that had a different config name than it equivalentclass type name

😱

In that case, thanks!

> So right now it is alot faster if the config alias and the class name alias matches, but we have no efficient way to check that in runtime.

What is even the use case for this I wonder?

TBH never understood this aspect and it certainly would help to get rid of it … _thinking out loud_

> Do you have any other suggestions or comments?

My understanding is that after your hard work this is getting into its final stages, so I'll start a more in-depth review (I hope soon).

crissi:
I will probably clean it up a bit and do the cache part in A new PR

Få Outlook til iOS<https://aka.ms/o0ukef>
Sendt: Friday, July 12, 2019 6:15:00 AM
Til: rebing/graphql-laravel
Cc: Christian; Author
Emne: Re: [rebing/graphql-laravel] [WIP] lazy load types (#394)


No, there was no types in the tests covering types that had a different config name than it equivalentclass type name

😱

In that case, thanks!

So right now it is alot faster if the config alias and the class name alias matches, but we have no efficient way to check that in runtime.

What is even the use case for this I wonder?

TBH never understood this aspect and it certainly would help to get rid of it … thinking out loud

Do you have any other suggestions or comments?

My understanding is that after your hard work this is getting into its final stages, so I'll start a more in-depth review (I hope soon).

mfn:
Sounds good to me!

crissi:
> > So right now it is alot faster if the config alias and the class name alias matches, but we have no efficient way to check that in runtime.
> 
> What is even the use case for this I wonder?
> 

Had the same thought. In my application I I have made it work with class names instead. So I'dont have to also remember type aliases.

image

mfn:
Interesting!

I made a similar approach but decided to for `TYPE` constants and use them when referencing the types:
image

crissi:
I am thinking maybe this typeloader should be a switch people can toggle on, if they are using types "correctly" by having the same alias in config and in names? Will make us not fight the underlying framework and make it just work (FAST). Will of cause need some docs to explain how it should be set up

mfn:
I don't want to jump the gun but I rather have one clear system than two to maintain.

Sorry I haven't yet found enough time for it, I haven't forgotten it!

mfn:
Closing in favor of https://github.com/rebing/graphql-laravel/pull/405

PR #402, opened 2019-07-11
fix: graphiql support multi-level schema

s51983:
graphiql only support the first route param. fixed.


mfn:
Unfortunately the fix at least for Lumen compatibility is incorrect.

The same reason it's actually broken in `\Rebing\GraphQL\GraphQLController::query` too.

This code (for the `query` action at least) was fixed in `master`.

Please:
- submit any fixes against `master` branch
- consider the new `\Rebing\GraphQL\GraphQLController::getRouteParameters` method
- please also consider the changed `graphiql.prefix` in the config which probably is wrong in master
- add a test (at least for the graphiql schema call, see `\Rebing\GraphQL\Tests\Unit\EndpointTest::testGetGraphqiQL` as blueprint example)

Would love to get a proper fix here, thank you!

PR #405, opened 2019-07-14
Introduce ability to lazy load types

crissi:
Enables lazy loading of, when types are used "correctly" by having the same alias in config and in names.

Will make us not fight the underlying framework and make it just work (FAST). Will of cause need some docs to explain how it should be set up.

I noticed about a 4 times speed up, testing on graphql-laravel-performancetest (did have to change all the type alias to match)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,9 +71,15 @@ install:
 script:
   - if [[ $STATIC_ANALYSER = 1 ]]; then composer phpstan; fi
   - if [[ $CODE_COVERAGE = 0 ]]; then vendor/bin/phpunit --colors=always --verbose; fi
-  - if [[ $CODE_COVERAGE = 1 ]]; then phpdbg -qrr vendor/bin/phpunit --colors=always --verbose --coverage-clover=coverage.xml; fi
+  - |
+    if [[ $CODE_COVERAGE = 1 ]]; then
+      phpdbg -qrr vendor/bin/phpunit --colors=always --verbose --coverage-clover=coverage.xml
+      bash <(curl -s https://codecov.io/bash) -cF lazyload-off
+    fi
+  - |
+    if [[ $CODE_COVERAGE = 1 ]]; then
+      TESTS_ENABLE_LAZYLOAD_TYPES=1 phpdbg -qrr vendor/bin/phpunit --colors=always --verbose --coverage-clover=coverage.xml
+      bash <(curl -s https://codecov.io/bash) -cF lazyload-on
+    fi
   - if [[ $INSTALL_LARAVEL = 1 ]]; then export -f travis_retry ; tests/integration-larvavel.sh; fi
   - if [[ $INSTALL_LUMEN = 1 ]]; then export -f travis_retry ; tests/integration-lumen.sh; fi
-
-after_success:
- - if [[ $CODE_COVERAGE = 1 ]]; then bash <(curl -s https://codecov.io/bash); fi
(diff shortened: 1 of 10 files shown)


mfn:
Is this the same as https://github.com/rebing/graphql-laravel/pull/394 but optional?

crissi:
> Is this the same as #394 but optional?

this one requires the types are used "correctly" by having the same alias in config and in names

and it can be turned on if people have the correct type setup (I rather like this than #394)

crissi:
> Is this the same as #394 but optional?

it is a replacement for the other one, so one of them should be closed

mfn:
I'm pretty excited about, thanks for your contribution!

Additionally I did some small style fixes; not necessarily because I prefer them but to keep the PR diff clean. Also some small code clarity improvements.

One thing which bothered me was, because due to the setting being `false` by default and basically only one test "testing for it", how can we ensure everything else works correctly when being enabled?

Long story short, I decided to simply run the unit tests twice; once with lazyload_types=false and once with true. The one test which fails when lazyload_types=true is simply skipped in this case.

Thanks!

mfn:
@crissi I'm thinking of making it default to `true` for v2.

My "gut feeling" is that aliasing is a minority thing and the majority would benefit from having it enabled.

What's your take on it?

mfn:
See https://github.com/rebing/graphql-laravel/pull/414

crissi:
I think that is good idea👍

Få Outlook til iOS<https://aka.ms/o0ukef>
Sendt: lørdag, juli 20, 2019 10:32 AM
Til: rebing/graphql-laravel
Cc: Christian; Mention
Emne: Re: [rebing/graphql-laravel] Introduce ability to lazy load types (#405)


@crissi<https://github.com/crissi> I'm thinking of making it default totrue for v2.

My "gut feeling" is that aliasing is a minority thing and the majority would benefit from having it enabled.

What's your take on it?

mfn:
@crissi I'm actual planning reverting the default of enabling it.

As was discovered in https://github.com/rebing/graphql-laravel/pull/415 (and https://github.com/rebing/graphql-laravel/pull/417 which was ultimately merged), it does also not work with pagination as it's creates an on demand type who name by design differs from the class; this is inherently incompatible.

I could live with the default as long as only aliasing on its own is affected, but pagination is much more useful in practice and that wouldn't be a good experience no matter how much the docs say about this (technically it's the same "problem" with aliasing).

crissi:
Probably a good idea until we find a solution

Få Outlook til iOS<https://aka.ms/o0ukef>
Sendt: Tuesday, July 23, 2019 11:16:17 AM
Cc: Christian <c2n@live.dk>; Mention <mention@noreply.github.com>
Emne: Re: [rebing/graphql-laravel] Introduce ability to lazy load types (#405)


@crissi<https://github.com/crissi> I'm actual planning reverting the default of enabling it.

As was discovered in #415<https://github.com/rebing/graphql-laravel/pull/415> (and #417<https://github.com/rebing/graphql-laravel/pull/417> which was ultimately merged), it does also not work with pagination as it's creates an on demand type who name by design differs from the class; this is inherently incompatible.

I could live with the default as long as only aliasing on its own is affected, but pagination is much more useful in practice and that wouldn't be a good experience no matter how much the docs say about this (technically it's the same "problem" with aliasing).

mfn:
Default has been set to false (again) => https://github.com/rebing/graphql-laravel/pull/422 😢

Issue #99, opened 2018-05-16
Pagination: associated model isn't set in the config


faistoiplaisir:
Hi,

I'll try to explain my problem, but it's not easy.

In the following query, I have a Producer model, and a relation ship to list all products associate with a producer.
 
```
query listProducerPaginate($limit: Int, $page: Int) {
    producers_pagination(limit: $limit, page: $page) {
        data {
          	products {
                    item_title
            }
        },
        total
}
```

I'm using an eager load request to load all products, with settings the $fields->getSelect() in the request:

```
return Producer::with($with)
    ->where($where)
    ->select($fields->getSelect())->paginate($args['limit'], ['*'], 'page', $args['page']);
```

This doesn't work because the producer primary key of the Producer model isn't automatically added.

In the SelectedField class, I saw you do it like that :

```
private static function getTableNameFromParentType($parentType)
{
    return isset($parentType->config['model']) ? app($parentType->config['model'])->getTable() : null;
}
```

It works fine, except for your Pagination mecanism because the PaginationType doesn't set the model in the config. Finally, at the runtime, the producer id isn't selected and the query doesn't work (sql crash).

I'll try to modify your pagination type constructor like that and it works fine:

```
public function __construct($typeName, $customName = null)
{
    // Get the graphql type for the paginate model
    $typeData = GraphQL::type($typeName);
    $name = $customName ?: $typeName . '_pagination';
    $config = [
        'name'  => $name,            
        'fields' => array_merge($this->getPaginationFields(), [
            'data' => [
                // Type the graphql type
                'type'      => GraphQLType::listOf($typeData),
                'resolve'   => function(LengthAwarePaginator $data) { return $data->getCollection();  },
            ],
        ])
    ];
    // If the model is set in the graphql type, we set it in the PaginationType
    if (isset($typeData->config['model'])) {
        $config['model'] = $typeData->config['model'];
    }
    parent::__construct($config);
}
```

I'm not sure about the side effect that can be introduce by that, do you think it's correct ? It's possible to add that on your side ?

rebing:
Can you show me your `ProducerType` class? Does it have a `'model'` property in `$attributes`?

faistoiplaisir:
Yes, there's a model in the ProducerType.

```
class ProducerType extends GraphQLType
{
    protected $attributes = [
        'name' => 'Producer',
        'description' => 'A type Producer',
        'model' => Producer::class
    ];

    public function fields()
    {
        return [
            'id' => [
[...]

PR #411, opened 2019-07-19
fixed selection of primary key in pagination type

Argentum88:
The problem is described here - #99


mfn:
Can you please make a PR against `master` and include tests showing your fix working?

Argentum88:
Ok.

PR #415, opened 2019-07-22
fixed selection of primary key in pagination type

Argentum88:
The problem is described here - #99


Argentum88:
Oh is some problem with lazy load of pagination type.

Argentum88:
I am disabled this functionality for my test

mfn:
Many thanks for your PR, I've merged it via https://github.com/rebing/graphql-laravel/pull/417 !

I had to adapt some smallish things but couldn't push to your PR (maybe it's disabled or maybe because of `master`).

Rest assured, I picked your commits so your authorship is kept 👍 

Thanks again!

PR #419, opened 2019-07-23
add multiple types of parameter resolution

sowork:
$args is parsed incorrectly when the parameter passed in is an object or listObject

```
{
    "query": "{users{id, name blogs(filter:{id:1, desc:\"hello\", ReviewInput:{id: 4}}, field5:[{id:2,desc:\"world\"},{id:3,desc:\"world\", ReviewInput:{id: 4}}]){id text comments{id desc} }}}"
}

$args:   result:
Array
(
    [filter] => Array
        (
            [id] => 1
            [desc] => hello
            [ReviewInput] => Array
                (
                    [id] => 4
                )

        )

    [field5] => Array
        (
            [0] => Array
                (
                    [id] => 2
                    [desc] => world
                )

            [1] => Array
                (
                    [id] => 3
                    [desc] => world
                    [ReviewInput] => Array
                        (
                            [id] => 4
                        )

                )

        )

)

```


sowork:
$args is parsed incorrectly when the parameter passed in is an object or listObject

```
{
    "query": "{users{id, name blogs(filter:{id:1, desc:\"hello\", ReviewInput:{id: 4}}, field5:[{id:2,desc:\"world\"},{id:3,desc:\"world\", ReviewInput:{id: 4}}]){id text comments{id desc} }}}"
}

$args:   result:
Array
(
    [filter] => Array
        (
            [id] => 1
            [desc] => hello
            [ReviewInput] => Array
                (
                    [id] => 4
                )

        )

    [field5] => Array
        (
            [0] => Array
                (
                    [id] => 2
                    [desc] => world
                )

            [1] => Array
                (
                    [id] => 3
                    [desc] => world
                    [ReviewInput] => Array
                        (
                            [id] => 4
                        )

                )

        )

)

```

sowork:
looks like something went wrong
```
Ignored error pattern /Access to an undefined property GraphQL\\Language\\AST\\ValueNode::\$value/ was not matched in reported errors.
```
what do I need to do?

mfn:
Please:
- don't format / change existing / unrelated code
- add tests!

Thank you!

sowork:
> don't format / change existing / unrelated code

I don't understand what you mean, can you say something specific?

src/Support/ResolveInfoFieldsAndArguments.php
```diff
@@ -132,15 +139,67 @@ private function foldSelectionSet(SelectionSetNode $selectionSet, int $descend):
...
+    private function getValue(ValueNode $value)
     {
-        $value = $argumentNode->value;
         if ($value instanceof VariableNode) {
             $variableName = $value->name->value;
 
             return $this->info->variableValues[$variableName] ?? null;
         }
+        // Scalar Types
+        if ($value instanceof IntValueNode || $value instanceof FloatValueNode || $value instanceof StringValueNode || $value instanceof BooleanValueNode || $value instanceof EnumValueNode) {
```
crissi: I think this could be shorten to: 

    if ($value instanceof ValueNode)

mfn:
> I don't understand what you mean, can you say something specific?

Please don't just remove the empty lines:
image

Also this change seems unrelated to any functionality in this PR:
image

But most importantly would be tests first.

src/Support/ResolveInfoFieldsAndArguments.php
sowork: Valuenode contains the following types

/**
export type ValueNode = VariableNode
| NullValueNode
| IntValueNode
| FloatValueNode
| StringValueNode
| BooleanValueNode
| EnumValueNode
| ListValueNode
| ObjectValueNode
 */

The parameter is ListValueNode or ObjectValueNode is not applicable.

sowork:
> > I don't understand what you mean, can you say something specific?
> 
> Please don't just remove the empty lines:
> image
> 
> Also this change seems unrelated to any functionality in this PR:
> image
> 
> But most importantly would be tests first.

This may be caused by my editor, I will fix it later and append a test

thank

sowork:
I added a unit test and it should be fine now.

src/Support/ResolveInfoFieldsAndArguments.php
```diff
@@ -132,15 +143,67 @@ private function foldSelectionSet(SelectionSetNode $selectionSet, int $descend):
...
+        // object Type
+        if ($value instanceof ObjectValueNode) {
+            return $this->getInputObjectValue($value);
+        }
+        // list object Type
+        if ($value instanceof ListValueNode) {
+            return $this->getInputListObjectValue($value);
+        }
+
+        throw new \Exception('Failed to resolve unknown ValueNode type');
```
mfn: If you don't add a specific exception, please use at least `RuntimeException` (and import it)

mfn: How can this be triggered with the current code?

Would like to see a test for this.

src/Support/ResolveInfoFieldsAndArguments.php
```diff
@@ -76,6 +85,7 @@ public function __construct(ResolveInfo $info)
      * @param  int  $depth  How many levels to include in output
      * @return array
      * @see \GraphQL\Type\Definition\ResolveInfo::getFieldSelection
+     * @throws \Exception
```
mfn: Please remove all the `@throws` you added in this PR

They are deliberately not added because it's a PITA to maintain them. Code changes, exception changes and before you notice it, it's out of date.

src/Support/ResolveInfoFieldsAndArguments.php
```diff
@@ -132,15 +143,67 @@ private function foldSelectionSet(SelectionSetNode $selectionSet, int $descend):
...
         if ($value instanceof VariableNode) {
             $variableName = $value->name->value;
 
             return $this->info->variableValues[$variableName] ?? null;
         }
+        // Scalar Types
+        if ($value instanceof IntValueNode || $value instanceof FloatValueNode || $value instanceof StringValueNode || $value instanceof BooleanValueNode || $value instanceof EnumValueNode) {
+            return $value->value;
+        }
+        // null Type
```
mfn: I think the code speaks for itself, remove the comment (but leave a blank line)

src/Support/ResolveInfoFieldsAndArguments.php
```diff
@@ -132,15 +143,67 @@ private function foldSelectionSet(SelectionSetNode $selectionSet, int $descend):
...
+        if ($value instanceof NullValueNode) {
+            return;
+        }
+        // object Type
```
mfn: I think the code speaks for itself, remove the comment (but leave a blank line)

src/Support/ResolveInfoFieldsAndArguments.php
```diff
@@ -132,15 +143,67 @@ private function foldSelectionSet(SelectionSetNode $selectionSet, int $descend):
...
+        if ($value instanceof ObjectValueNode) {
+            return $this->getInputObjectValue($value);
+        }
+        // list object Type
```
mfn: I think the code speaks for itself, remove the comment (but leave a blank line)

src/Support/ResolveInfoFieldsAndArguments.php
```diff
@@ -132,15 +143,67 @@ private function foldSelectionSet(SelectionSetNode $selectionSet, int $descend):
...
+        }
+
+        throw new \Exception('Failed to resolve unknown ValueNode type');
+    }
+
+    /**
+     * @param ObjectValueNode $objectValueNode
+     * @return array
+     * @throws \Exception
+     */
```
mfn: Remove the docblock please (throws should not be used and the rest does not add anything)

src/Support/ResolveInfoFieldsAndArguments.php
```diff
@@ -132,15 +143,67 @@ private function foldSelectionSet(SelectionSetNode $selectionSet, int $descend):
...
+    private function getInputObjectValue(ObjectValueNode $objectValueNode)
```
mfn: add return type `:array`

src/Support/ResolveInfoFieldsAndArguments.php
```diff
@@ -97,6 +107,7 @@ public function getFieldsAndArgumentsSelection(int $depth = 0): array
      * @param  int  $descend
      * @return array
      * @see \GraphQL\Type\Definition\ResolveInfo::foldSelectionSet
+     * @throws \Exception
```
mfn: Remove

src/Support/ResolveInfoFieldsAndArguments.php
```diff
@@ -132,15 +143,67 @@ private function foldSelectionSet(SelectionSetNode $selectionSet, int $descend):
         return $fields;
     }
 
-    private function getValue(ArgumentNode $argumentNode)
+    /**
+     * @param ValueNode $value
+     * @return mixed
+     * @throws \Exception
+     */
```
mfn: Remoe docblock

src/Support/ResolveInfoFieldsAndArguments.php
```diff
@@ -132,15 +143,67 @@ private function foldSelectionSet(SelectionSetNode $selectionSet, int $descend):
...
+        }
+
+        return $value;
+    }
+
+    /**
+     * @param ListValueNode $listValueNode
+     * @return array
+     * @throws \Exception
+     */
```
mfn: remove phpdoc

src/Support/ResolveInfoFieldsAndArguments.php
```diff
@@ -132,15 +143,67 @@ private function foldSelectionSet(SelectionSetNode $selectionSet, int $descend):
...
+    private function getInputListObjectValue(ListValueNode $listValueNode)
```
mfn: add return type decl `:array`

src/Support/ResolveInfoFieldsAndArguments.php
```diff
@@ -132,15 +143,67 @@ private function foldSelectionSet(SelectionSetNode $selectionSet, int $descend):
...
 
             return $this->info->variableValues[$variableName] ?? null;
         }
+        // Scalar Types
+        if ($value instanceof IntValueNode || $value instanceof FloatValueNode || $value instanceof StringValueNode || $value instanceof BooleanValueNode || $value instanceof EnumValueNode) {
+            return $value->value;
+        }
+        // null Type
+        if ($value instanceof NullValueNode) {
+            return;
```
mfn: Shouldn't this be more explicit and actually return `null`?

src/Support/ResolveInfoFieldsAndArguments.php
```diff
@@ -132,15 +143,67 @@ private function foldSelectionSet(SelectionSetNode $selectionSet, int $descend):
...
+    /**
+     * @param ObjectValueNode $objectValueNode
+     * @return array
+     * @throws \Exception
+     */
+    private function getInputObjectValue(ObjectValueNode $objectValueNode)
+    {
+        $value = [];
+        foreach ($objectValueNode->fields->getIterator() as $item) {
+            if ($item instanceof ObjectFieldNode) {
```
mfn: Not sure about this.

The code makes it look like `$item` could be something else. But reading `\GraphQL\Language\AST\ObjectValueNode::$fields` it's clear this is not intended.

How can we improve this? Should we remove the check? Or add an `} else {` and throw an exception?

Because as it is right now, if that case ever would happen, we silently would carry on.

phpstan.neon.dist
```diff
@@ -42,8 +42,6 @@ parameters:
         - '/Cannot call method set\(\) on Illuminate\\Config\\Repository\|null/'
         - '/Parameter #4 \$currentPage of class Illuminate\\Pagination\\LengthAwarePaginator constructor expects int\|null, float\|int given/'
         - '/Parameter #1 \$offset of method Illuminate\\Support\\Collection::slice\(\) expects int, float\|int given/'
-        # \Rebing\GraphQL\Support\ResolveInfoFieldsAndArguments::getValue
-        - '/Access to an undefined property GraphQL\\Language\\AST\\ValueNode::\$value/'
```
mfn: This pretty cool and a very clear indicator that the current code was broken :/ Great 👍

src/Support/ResolveInfoFieldsAndArguments.php
```diff
@@ -132,15 +143,67 @@ private function foldSelectionSet(SelectionSetNode $selectionSet, int $descend):
...
+    private function getValue(ValueNode $value)
     {
-        $value = $argumentNode->value;
         if ($value instanceof VariableNode) {
             $variableName = $value->name->value;
 
             return $this->info->variableValues[$variableName] ?? null;
         }
+        // Scalar Types
+        if ($value instanceof IntValueNode || $value instanceof FloatValueNode || $value instanceof StringValueNode || $value instanceof BooleanValueNode || $value instanceof EnumValueNode) {
```
mfn: Please break the lines: there are so many ORs, they really need to be better readable.

For simplicity, I suggest each one gets its own line:
```php
        if ($value instanceof IntValueNode ||
            $value instanceof FloatValueNode ||
…
```

src/Support/ResolveInfoFieldsAndArguments.php
mfn: @sowork @crissi should we change the order of conditions?

1. check for ListValueNode and ObjectVakueNode
2. Check for ValueNode

This way we would not be concerned anymore about all the details as they effectively work the same.

Another approach for 2.: we effectively treat the "fallthrough" in this method as always being `ValueNode` and return `$argumentNode->value` as it currently is?

Would that make sense?

sowork: Theoretically feasible, but there is also a nullValueNode

crissi: not sure

but checking for the ListValueNode and ObjectVakueNode and nullValueNode first

and then fallthough to just get the value might work. So custom scalar types is also considered, not sure the current implementation take this in account.

mfn: > but there is also a nullValueNode

Right! Forgot to list it, but that would apply for it too.

> So custom scalar types 

Oh, interesting. Didn't think about it and currently don't know what kind of node it exactly is.

We need a test for this too. 😄

src/Support/ResolveInfoFieldsAndArguments.php
sowork: 1. Currently, item only handles objects of type ObjectFieldNode
2. If the current item object contains other scalar types or listNode and so on, the resolution continues recursively

src/Support/ResolveInfoFieldsAndArguments.php
sowork: Originally I returned null, but `styleci` changed to this way by default.

src/Support/ResolveInfoFieldsAndArguments.php
sowork: > custom scalar types

for custom scalar types will still be converted to a  `ValueNode`

Issue #421, opened 2019-07-23
Custom query is not called on interface's relation field


EdwinDayot:
### Versions:
- graphql-laravel Version: 2.0.0-RC4
- Laravel Version: 5.8
- PHP Version: 7.2

### Description:

`src/Rebing/GraphQL/Support/SelectFields.php`
```php
// Line 160
if (isset($parentType->config['model'])) {
```

Because of this line, an `Interface` (and the `Type`s implementing it) cannot have a  field using a `customQuery`.

This is also true for a `Union`.

This `if` statement probably should have an `or` statement for `Interface`s and `Union`s

### Steps To Reproduce:

- Create an interface containing a field with a custom query

EdwinDayot:
After a deeper investigation, adding a `or` statement wouldn't be enough since the `model` itself is required to get the `relation`. Still, our `Type`s implementing the `Interface` have access to the `model` property.

mfn:
You mean there's a clash in case a field has the _literal_ name `customQuery`?

It would be great to get some tests showing this problem!

EdwinDayot:
No, I mean something like this:

```php
class ObjectInterfaceType extends InterfaceType
{
    /**
     * @var array
     */
    protected $attributes = [
        'name' => 'ObjectInterfaceType',
        'description' => 'A type'
    ];

    /**
     * @return array
     */
    public function fields(): array
    {
        return [
            'id' => [
                'type' => Type::nonNull(Type::id()),
                'description' => 'The id of the object'
            ],
            'statistics' => [
                'type' => Type::listOf(GraphQL::type('statistic')),
                'args' => [
                    'group' => ['name' => 'group', 'type' => Type::string()]
                ],
                'query' => function (array $args, $query) {
                    // Never reached
                }
            ]
        ];
    }


    /**
     * @param $root
     * @return mixed
     */
    public function resolveType($root)
    {
        if ($root instanceof Entity) {
            return GraphQL::type('someType');
        }
    }
}
```

```php
class SomeType extends GraphQLType
{
    protected $attributes = [
        'name' => 'SomeType',
        'description' => 'A type',
        'model' => Entity::class,
    ];

    public function fields(): array
    {
        $interface = GraphQL::type('objectInterface');
        $result =
            [
                'statistics' => [
                    'type' => Type::listOf(GraphQL::type('someStatisticType')),
                    'query' => function (array $args, $query) {
                        // Never reached
                    }
                ]
            ] + $interface->getFields();;
[...]

PR #426, opened 2019-07-25
Added a test for custom query on a relation in an interface

EdwinDayot:
See https://github.com/rebing/graphql-laravel/issues/421

### DIFF
--- a/tests/Database/SelectFields/InterfaceTests/InterfaceTest.php
+++ b/tests/Database/SelectFields/InterfaceTests/InterfaceTest.php
@@ -6,6 +6,7 @@ namespace Rebing\GraphQL\Tests\Database\SelectFields\InterfaceTests;
 
 use Rebing\GraphQL\Tests\TestCaseDatabase;
 use Rebing\GraphQL\Tests\Support\Models\Post;
+use Rebing\GraphQL\Tests\Support\Models\Comment;
 use Rebing\GraphQL\Tests\Support\Traits\SqlAssertionTrait;
 
 class InterfaceTest extends TestCaseDatabase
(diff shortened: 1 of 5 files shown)


mfn:
Can't consider merging until everything is green

EdwinDayot:
Sure, it will require to take in consideration each `Type` implementing an `Interface` to take the right `model` and `relation`. I was thinking about some kind of `Closure` taking `model` as parameter once it is required, but I really don't know how to implement this. As I've no idea how to solve this at the moment, would you recommend a temporary solution to make this mergeable ?

I could write tests returning green, but I'd prefer it not to be forgotten

tests/Database/SelectFields/InterfaceTests/InterfaceTest.php
```diff
@@ -47,6 +48,58 @@ public function testGeneratedSqlQuery(): void
         $this->assertSame($expectedResult, $result);
     }
 
+    public function testGeneratedRelationSqlQuery()
```
mfn: Please always add the return type (`: void` in this case), like we do it everywhere, thank you!

tests/Database/SelectFields/InterfaceTests/InterfaceTest.php
```diff
@@ -47,6 +48,58 @@ public function testGeneratedSqlQuery(): void
...
+                'title' => 'Title of the comment',
+                'post_id' => $post->id,
+            ]);
+
+        $graphql = <<<'GRAPHQL'
+{
+  exampleInterfaceQuery {
+    id
+    title
+    exampleRelation{
```
mfn: Space ` ` before `{`

mfn:
> I could write tests returning green, but I'd prefer it not to be forgotten

Oh, I see.

Well, a red test is unfortunately not the right approach to "keep track" of it. First: can't merge PRs not green, a good precaution.

Second: future PRs may or may not break stuff and then it's harder to segregate these issues.

Having a test showing "something is broke" or "not working" is totally fine! You just have to use other means to remind about, like an open bug report referencing this PR, etc.

mfn:
Thank you very much for the test 👍

PR #429, opened 2019-07-25
[TESTS] Remove code coverage via codecov

mfn:
- in practice I didn't find it very useful
- more often then not and especially recently it simply didn't work,
  timed out when trying to upload data

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ cache:
 
 env:
   global:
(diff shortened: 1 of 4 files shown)


mfn:
Dang, removing the code coverage part and thus adapting the shell exposed that when lazy loading is enabled, the test suite actually fails due to the recent supposedly compatible pagination fixes

PR #431, opened 2019-07-27
Remove Fluent dependency on core `Field`

mfn:
Same reasoning as https://github.com/rebing/graphql-laravel/pull/389

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,7 +53,8 @@ CHANGELOG
   - Previously the initial query arguments would be used everywhere
 
 ### Removed
-- Removed `\Fluent` dependency on `\Rebing\GraphQL\Support\Type`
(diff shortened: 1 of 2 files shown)


mfn:
This will also _slightly_ decrease memory usage on large query/mutations, like we've on https://github.com/dnyg/graphql-laravel-performancetest .

After adding the following to the `GraphqlQuery.php` in that repo and measure before/after:
```diff
+        echo 'Memory memory_get_usage: ' .  memory_get_usage() . PHP_EOL;
+        echo 'Memory memory_get_peak_usage: ' .  memory_get_peak_usage() . PHP_EOL;
```

#### Before:
```
Memory memory_get_usage: 84_186_888
Memory memory_get_peak_usage: 84_230_000
```

#### After
```
Memory memory_get_usage: 84_015_968
Memory memory_get_peak_usage: 84_059_080
```

The usage is stable for every iteration. Probably not the primary reason to do this, but I like to cut down on "unintended" dependencies like the `\Fluent` class which otherwise doesn't bring much to the table anywhere here.

PR #432, opened 2019-07-27
Remove unused __isset()

mfn:
Let's see if this is really "needed"


mfn:
Really couldn't find any use case in the current codebase.

PR #433, opened 2019-07-27
Fix pagination for lazy loading

crissi:


mfn:
🎉

src/GraphQL.php
```diff
@@ -332,11 +332,13 @@ protected function clearTypeInstances(): void
 
     public function paginate(string $typeName, string $customName = null): Type
     {
-        $name = $customName ?: $typeName.'_pagination';
+        $name = $customName ?: $typeName.'Pagination';
```
mfn: Could this have any kind of compatibility impact?

mfn: And, curious, I compared this to your other PR and unless I made a mistake, that's the only difference I could see. Why does that fix it? 🤷‍♀️

crissi: take a look in the PaginationType, it is the same way the name is constructed

crissi: > Could this have any kind of compatibility impact?

I don't think so

PR #437, opened 2019-08-02
handling wrapped types

w1ldzer0:
Now, if types wrapped smth like, `Type::nonNull` or `Type::listOf`, it wouldn't handle. This commit fixes it

### DIFF
--- a/src/Support/SelectFields.php
+++ b/src/Support/SelectFields.php
@@ -207,7 +207,12 @@ class SelectFields
 
                         $with[$relationsKey] = self::getSelectableFieldsAndRelations($queryArgs, $field, $newParentType, $customQuery, false);
                     } else {
-                        self::handleFields($queryArgs, $field, $fieldObject->config['type'], $select, $with);
+                        if (method_exists($fieldObject->getType(), 'getWrappedType')) {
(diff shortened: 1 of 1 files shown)


src/Support/SelectFields.php
```diff
@@ -207,7 +207,12 @@ protected static function handleFields(array $queryArgs, array $requestedFields,
 
                         $with[$relationsKey] = self::getSelectableFieldsAndRelations($queryArgs, $field, $newParentType, $customQuery, false);
                     } else {
-                        self::handleFields($queryArgs, $field, $fieldObject->config['type'], $select, $with);
+                        if (method_exists($fieldObject->getType(), 'getWrappedType')) {
```
mfn: Why would you check for the method and not if its an instance of `WrappingType`? See existing use for this in the very same class.

I think in this case I would prefer a ternary operator, something like:
```php
$type = … instanceof …
    ? …
    : …;
```

mfn:
ping @w1ldzer0

mfn:
ping @w1ldzer0 

I intend to close this PR soon due to missing feedback

mfn:
Closing due to missing feedback

PR #441, opened 2019-08-10
Updated readme to reflect deprecation of InputType & EnumType

maagiline:


mfn:
🤔

How did you come to this conclusion?

It's actually the other way around (and as documented in the readme), using the properties is deprecated as can be seen in the following places:
- https://github.com/rebing/graphql-laravel/blob/34492b7df6509d8f31c6eb419a8b974ab46a88be/src/Support/Type.php#L24
- https://github.com/rebing/graphql-laravel/blob/34492b7df6509d8f31c6eb419a8b974ab46a88be/src/Support/Type.php#L31

maagiline:
Aha, interesting. I somehow managed to install v1.24 instead of 2.x, but relied on 2.x-s docs. Hence the huge confusion. My bad!

PR #447, opened 2019-08-15
update routes.php

MonsterDong:
multi level path for route where

### DIFF
--- a/src/routes.php
+++ b/src/routes.php
@@ -76,7 +76,18 @@ $router->group(array_merge([
                     );
 
                     if (! Helpers::isLumen()) {
-                        $route->where($name, $name);
+                        $multiLevelPath = explode('/', $name);
+                        if (count($multiLevelPath) > 1) {
+                            
+                            $where = [];
+                            foreach ($multiLevelPath as $multiName) {
+                                $where[$multiName] = $multiName;
+                            }
+
+                            $route->where($where);
+                        } else {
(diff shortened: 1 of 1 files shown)


mfn:
Please explain in more detail what this is about and also add some tests

MonsterDong:
> Please explain in more detail what this is about and also add some tests

image
image

I defined two schemas, user/me and merchant/me. Merchant /me configured the middleware mustBindPhone(which must be tied with the phone to access it), but when I visited user/me schemas, I chose the middleware of merchant/me schemas

MonsterDong:
Versions:
graphql-laravel Version: v1.24
Laravel Version: 5.8
PHP Version: 7.2

mfn:
Is the problem that the `/` in the schema name confuses the routing?

MonsterDong:
> Is the problem that the `/` in the schema name confuses the routing?
Yes, this problem causes $router->where() not to take effect

mfn:
It seems you added it to the "if not lumen" condition; any idea how both can be supported?

MonsterDong:
> It seems you added it to the "if not lumen" condition; any idea how both can be supported?
In the lumen
image


In the code

image

So there's no problem with lumen

mfn:
ping @MonsterDong

mfn:
Closing due to missing feedback (i.e. tests, in this case; sorry but can't accept such changes without them!)

Issue #444, opened 2019-08-13
Get Illuminate\Http\Request within a Mutation?


ajanini:
### Versions:
- graphql-laravel Version: 1.23.0
- Laravel/Lumen Version: 5.7.28
- PHP Version: 7.2.1

### Question:

How do I get the original `\Illuminate\Http\Request` within a GraphQL *mutation*?

Thanks

mfn:
Although the library receives the `Request` in `\Rebing\GraphQL\GraphQLController::query`, it does not pass it on.

What this means is: the global `request()` helper is your best option at the moment.

mfn:
Although there's the workaround I mentioned (others might say it's not event a "workaround" but correct), it think it should be possible to carry it along with the resolvers 3rd argument, the `$context`.

However for this to work the following is necessary:
- a custom controller extending `\Rebing\GraphQL\GraphQLController`
- override `queryContext`
- from it return what you want (either the `Request` directly or, preferable, some wrapper object where you put the request into it)

I implemented this suggestion in https://github.com/rebing/graphql-laravel/pull/450 : it annoys me that it's a) a breaking change b) adds yet-another-parameter to `queryContext` but I don't see a better way right now. Please give it a try.

ajanini:
Thank you so much, @mfn.

I went with your first suggestion, using the global `request()` and it worked quite nicely for my needs.

I really appreciate your help.

PR #450, opened 2019-08-15
Add the current `\Illuminate\Http\Request` as 4th parameter to `\Rebing\GraphQL\GraphQLController::queryContext`

mfn:
This allows creating a context with the current request without having to inject it via `request()` or other means when needed in a resolver.

This, unfortunately, is a breaking change for everyone currently implementing a custom context because the signature of `queryContext()` changed and needs to be adapted.

Fixes https://github.com/rebing/graphql-laravel/issues/444


mfn:
Closing this for now, I'm not super happy with the approach

PR #451, opened 2019-08-16
Separate Larastan config

szepeviktor:
@mfn What do you think?


phpstan.neon.dist
```diff
@@ -1,23 +1,13 @@
+includes:
+    - larastan.neon
```
mfn: I'm wondering, can't we just reference the config _shipped_ with larastan that way? 🤷‍♀️

szepeviktor: Feel free to experiment with it. Actually I promote phpstan, I'm not a graphql user.

szepeviktor: `vendor/nunomaduro/larastan/extension.neon`

szepeviktor: I do not know how to _un-ignore_ errors.

mfn: > I do not know how to _un-ignore_ errors.

Ahh. I guess that what the reason in the first place why I assimilated it.

szepeviktor:
You're welcome.

PR #452, opened 2019-08-16
allow alias to be a callback

crissi:
a test actually existed for this case but just repeated the previous test

### DIFF
--- a/src/Support/SelectFields.php
+++ b/src/Support/SelectFields.php
@@ -213,7 +213,7 @@ class SelectFields
                 // Select
                 else {
                     $key = isset($fieldObject->config['alias'])
-                        ? $fieldObject->config['alias']
+                        ? value($fieldObject->config['alias'])
                         : $key;
 
                     self::addFieldToSelect($key, $select, $parentTable, false);
--- a/src/Support/SelectFields.php
+++ b/src/Support/SelectFields.php
@@ -213,8 +213,9 @@ class SelectFields
                 // Select
                 else {
                     $key = isset($fieldObject->config['alias'])
-                        ? value($fieldObject->config['alias'])
+                        ? $fieldObject->config['alias']
                         : $key;
+                    $key = $key instanceof Closure ? $key() : $key;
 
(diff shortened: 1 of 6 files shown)


src/Support/SelectFields.php
```diff
@@ -213,7 +213,7 @@ protected static function handleFields(array $queryArgs, array $requestedFields,
                 // Select
                 else {
                     $key = isset($fieldObject->config['alias'])
-                        ? $fieldObject->config['alias']
+                        ? value($fieldObject->config['alias'])
```
mfn: I know these global helpers are super convenient, but the same reason they're global I despise them.

In this case this should be replaced with inline code doing the same.

tests/Database/SelectFieldsTest.php
```diff
@@ -318,24 +330,27 @@ public function testWithSelectFieldsAndModelAndCallbackSqlAlias(): void
...
+                    'commentsLastMonth' => 1,
                 ],
             ],
         ];
 
         $this->assertEquals($response->getStatusCode(), 200);
         $this->assertEquals($expectedResult, $response->json());
+
+        // clear the mock
+        Carbon::setTestNow();
```
mfn: Now necessary, Laravels (and Orchestras) `tearDown` takes care of it

tests/Database/SelectFieldsTest.php
```diff
@@ -295,17 +298,26 @@ public function testWithSelectFieldsAndModelAndCallbackSqlAlias(): void
             'title' => 'Description of the post',
         ]);
 
-        factory(Comment::class, 2)
+        Carbon::setTestNow('2018-01-01');
+
+        factory(Comment::class)
             ->create([
                 'post_id' => $post->id,
+                'created_at' => new DateTime('2000-01-01'),
```
mfn: I'd rather stay in "Carbon-town" everywhere.

mfn:
> a test actually existed for this case but just repeated the previous test

I remember, guess I was too enthusiastic in https://github.com/rebing/graphql-laravel/pull/356 ;)

PR #483, opened 2019-09-18
Fix validation of list of scalar types

crissi:
this would actually apply ex. the **email** rule for each element in a list of strings.

The downside of this is that you can no longer create a rule for the whole array. 
Maybe a kind of Rule wrapper could be created to make a specific rule work for the whole list...

I am not sure this is the right way to fix this actually, maybe it should be fixed in userland code..?!?


crissi:
closing, i think this might make more sense to implement i your own codebase...

PR #485, opened 2019-10-05
Added a function to add custom wrap for the queries or mutations

albertcito:
Now you can create my own wrapper like pagination to inject more options to te resolve function in my GraphQL queries or mutations.


albertcito:
I have done the test. I have two questions before to push the code.

1.  Do you have some command to verify the style code in my local environment? Something like that ` phpcs --standard=phpcs.xml --extensions=php src`.
2. About "new public method to the facade". That means add a new function in `src/Support/Facades/GraphQL.php` like this one: 

```php
protected static function wrapType()
{
    return 'graphql';
}
```

mfn:
> 1. Do you have some command to verify the style code in my local environment? Something like that ` phpcs --standard=phpcs.xml --extensions=php src`.

Unfortunately no, StyleCI entirely works "in the cloud"; so only after a push you get back the information via Github. The "best" you can do then, instead of manually fixing it, is to simply download the diff and apply it, push again (or, as I usually do it, rebase)



> 2. About "new public method to the facade". That means add a new function in `src/Support/Facades/GraphQL.php` like this one:

No I meant adding it to the phpdoc in https://github.com/rebing/graphql-laravel/blob/e735562051cf1441664a99b493560d6dc47fdc0b/src/Support/Facades/GraphQL.php#L14-L26

mfn:
@albertcito the implementation and the tests look good on a high level, but I'm hesitant to merge because there are a lot of changes in the PR which are a) unrelated and b) don't make much sense.

For example, in your PR a lot of tests from the `InterfaceTests` namespace appear as added: that's strange because they are already in master. Something is odd here.

Can you solve this, make it so the PR only includes _your actual_ changes?

albertcito:
Yes, I'll create a new PR. Unfortunately I merged it and something wrong happened.

albertcito:
This PR was closed, because the changes will be added in the PR #496

PR #486, opened 2019-10-08
Fixed the custom query not being handled by interface's relations

EdwinDayot:
Fixes the #421 

An `Interface` requires a `public function types` returning an array of `GraphQL::type()` if it contains a relation with a custom `query`

Still, I'm not a huge fan of this method, so every new idea is appreciated!

### DIFF
--- a/src/Support/InterfaceType.php
+++ b/src/Support/InterfaceType.php
@@ -25,6 +25,21 @@ abstract class InterfaceType extends Type
         };
     }
 
+    protected function getTypesResolver(): ?Closure
+    {
+        if (! method_exists($this, 'types')) {
+            return null;
+        }
+
+        $resolver = [$this, 'types'];
+
+        return function () use ($resolver): array {
+            $args = func_get_args();
+
+            return call_user_func_array($resolver, $args);
+        };
+    }
+
     /**
      * Get the attributes from the container.
      *
@@ -34,9 +49,14 @@ abstract class InterfaceType extends Type
     {
         $attributes = parent::getAttributes();
 
-        $resolver = $this->getTypeResolver();
-        if ($resolver) {
-            $attributes['resolveType'] = $resolver;
+        $resolverType = $this->getTypeResolver();
+        if ($resolverType) {
+            $attributes['resolveType'] = $resolverType;
+        }
+
+        $resolverTypes = $this->getTypesResolver();
+        if ($resolverTypes) {
+            $attributes['types'] = $resolverTypes;
         }
 
         return $attributes;
--- a/src/Support/SelectFields.php
+++ b/src/Support/SelectFields.php
@@ -17,7 +17,9 @@ use Illuminate\Database\Eloquent\Relations\HasOne;
 use Illuminate\Database\Eloquent\Relations\MorphMany;
 use Illuminate\Database\Eloquent\Relations\MorphOne;
 use Illuminate\Database\Eloquent\Relations\MorphTo;
+use Illuminate\Database\Eloquent\Relations\Relation;
 use Illuminate\Database\Query\Expression;
+use Illuminate\Support\Arr;
 use RuntimeException;
 
 class SelectFields
@@ -114,6 +116,16 @@ class SelectFields
         };
     }
 
+    private static function getTableNameFromParentType(GraphqlType $parentType): ?string
+    {
+        return isset($parentType->config['model']) ? app($parentType->config['model'])->getTable() : null;
+    }
+
+    private static function getPrimaryKeyFromParentType(GraphqlType $parentType): ?string
+    {
+        return isset($parentType->config['model']) ? app($parentType->config['model'])->getKeyName() : null;
+    }
+
     /**
      * Get the selects and withs from the given fields
      * and recurse if necessary.
@@ -159,6 +171,12 @@ class SelectFields
                 continue;
             }
 
+            $parentTypeUnwrapped = $parentType;
+
+            if ($parentTypeUnwrapped instanceof WrappingType) {
+                $parentTypeUnwrapped = $parentTypeUnwrapped->getWrappedType(true);
+            }
+
             // First check if the field is even accessible
             $canSelect = self::validateField($fieldObject, $queryArgs);
             if ($canSelect === true) {
@@ -170,7 +188,10 @@ class SelectFields
 
                 // Pagination
                 if (is_a($parentType, config('graphql.pagination_type', PaginationType::class))) {
-                    self::handleFields($queryArgs, $field, $fieldObject->config['type']->getWrappedType(), $select, $with, $ctx);
+                    /* @var GraphqlType $fieldType */
+                    $fieldType = $fieldObject->config['type'];
+                    self::handleFields($queryArgs, $field, $fieldType->getWrappedType(), $select,
+                        $with, $ctx);
                 }
                 // With
                 elseif (is_array($field['fields']) && $queryable) {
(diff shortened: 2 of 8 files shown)


EdwinDayot:
Found out that interfaces public function types were useless

mfn:
Phu, the changes seem quite complex… I mean, on one side it looks like you just moved stuff around into a dedicated method, but hard to say.

Btw, this one test is failing; what is your test expectation here?

I'm a bit unsure if this change / added complexity is worth it, in its current form (no better suggestion though yet).

EdwinDayot:
The changes are pretty simple actually. Instead of executing all the code before any database request, in the case of an interface's relation, it is executed on the relation's query, as we don't know what model will be requested. Maybe this solution is better than nothing, and if anyone complains about the complexity, or anything, maybe someone will PR something else?

Anyway, if anyone has something better in mind...

mfn:
Ok, thx for the info.

I will take a closer look for the review, but that might take a while. If you want to add more details to the source changes (i.e. via github inline comments) to guide the reviews, that would be much appreciated.

src/Support/SelectFields.php
```diff
@@ -180,49 +188,17 @@ protected static function handleFields(
...
-                        }
-                        // If 'HasMany', then add it in the 'with'
-                        elseif ((is_a($relation, HasMany::class) || is_a($relation, MorphMany::class) || is_a($relation, HasOne::class) || is_a($relation, MorphOne::class))
-                            && ! array_key_exists($foreignKey, $field)) {
-                            $segments = explode('.', $foreignKey);
-                            $foreignKey = end($segments);
-                            if (! array_key_exists($foreignKey, $field)) {
-                                $field['fields'][$foreignKey] = self::FOREIGN_KEY;
-                            }
-                        }
```
EdwinDayot: Moved to `addForeignKey`

src/Support/InterfaceType.php
```diff
@@ -39,6 +54,11 @@ public function getAttributes(): array
             $attributes['resolveType'] = $resolver;
         }
 
+        $resolver = $this->getTypesResolver();
+        if ($resolver) {
+            $attributes['types'] = $resolver;
+        }
+
```
EdwinDayot: We need to add a method in each `Interface`: `types` returning an array of `Type`s. This way, we know what `Type`s to expect while executing our relation's query.

src/Support/SelectFields.php
```diff
@@ -390,6 +366,116 @@ private static function isMongodbInstance(GraphqlType $parentType): bool
...
+
+            self::addAlwaysFields($fieldObject, $field, $parentTable, true);
+
+            if ($newParentType instanceof WrappingType) {
+                $newParentType = $newParentType->getWrappedType(true);
+            }
+
+            if (isset($newParentType->config['types'])) {
+                foreach ($newParentType->config['types']() as $type) {
+                    if (app($type->config['model'])->getTable() === $query->getRelated()->getTable()) {
```
EdwinDayot: For each types of the interface, we are fetching the table of the model. If it is the same table as the one we are currently querying from, then we should build our custom query, and execute it. (`return $callable($query);`)

src/Support/SelectFields.php
```diff
@@ -390,6 +366,116 @@ private static function isMongodbInstance(GraphqlType $parentType): bool
...
+                        $callable = self::getSelectableFieldsAndRelations($queryArgs, $field, $type, $customQuery, false, $ctx, false);
+
+                        return $callable($query);
+                    }
+                }
+            }
+            /** @var callable $callable */
+            $callable = self::getSelectableFieldsAndRelations($queryArgs, $field, $newParentType, $customQuery, false, $ctx, false);
+
+            return $callable($query);
```
EdwinDayot: Otherwise, we should call the custom query without giving it the relation's type

tests/Database/SelectFields/InterfaceTests/InterfaceTest.php
```diff
@@ -78,7 +78,7 @@ public function testGeneratedRelationSqlQuery(): void
 
         $this->assertSqlQueries(<<<'SQL'
 select "id", "title" from "posts";
-select * from "comments" where "comments"."post_id" = ? and "comments"."post_id" is not null order by "comments"."id" asc;
+select "comments"."title", "comments"."post_id" from "comments" where "comments"."post_id" in (?) and "id" >= ? order by "comments"."id" asc;
```
EdwinDayot: It results in the fact that the interface gives the fields that should be selected for the relation instead of selecting `*`, and it allows a custom query to be passed to this relation's query.

src/Support/SelectFields.php
```diff
@@ -390,6 +366,116 @@ private static function isMongodbInstance(GraphqlType $parentType): bool
...
+        array &$with,
+        $ctx,
+        $fieldObject,
+        $key,
+        $customQuery
+    ) {
+        $relationsKey = Arr::get($fieldObject->config, 'alias', $key);
+
+        $with[$relationsKey] = function ($query) use ($queryArgs, $field, $parentType, &$select, $ctx, $customQuery, $key, $fieldObject) {
+            $parentTable = self::isMongodbInstance($parentType) ? null : self::getTableNameFromParentType($parentType);
```
EdwinDayot: This part is executed when a `Query` calls the `->with()` method of the model.

See it as a query like so:
```php
MyModel::with(function ($query) {
  // We are here
})->get();
```

mfn:
Thanks for your effort _and_ your patience so far, really appreciated.

I took a longer look today and, whilst not easy, the approach seems 👍 to me!

But my feeling is that it's lacking tests. Only one existing test adapt (which I remember was added in advance for exactly this moment) isn't much.
`\Rebing\GraphQL\Support\SelectFields::handleInterfaceFields` also seems to have some uncovered parts in the area of `if (isset($newParentType->config['types'])) {`.

My suggestion would be these steps in order:
- please add much tests as you can think of / possible
- invite others to also give feedback; currently thinking about @crissi as being also quite active lately
- final review and 🤞 merge

How does that sound for your?

PS: without getting ahead, `handleInterfaceFields` seems like it could use more specific PHP types on the function arguments?

Thanks!

mfn:
@EdwinDayot I see so often that you've to make micro commits for the StyleCI violations: it should be _rather_ easy => just click on the "Details" and download / apply the patch.

Good news: we want to switch to something else which allows devs to simply run php cs fixer locally, but it's not yet done => https://github.com/rebing/graphql-laravel/pull/502

EdwinDayot:
Could not find the issue with Laravel 5.5 for the moment, if anyone has an idea...

EdwinDayot:
> @EdwinDayot I see so often that you've to make micro commits for the StyleCI violations: it should be _rather_ easy => just click on the "Details" and download / apply the patch.
> 
> Good news: we want to switch to something else which allows devs to simply run php cs fixer locally, but it's not yet done => #502

Yeah, that's what I am doing. I wasn't aware of this feature at the beginning, but now, I'm using it :)

EdwinDayot:
Help wanted

mfn:
Ops, the StyleCI integration is still active. On it!

tests/Database/SelectFields/InterfaceTests/LikableInterfaceType.php
```diff
@@ -24,7 +33,12 @@ public function fields(): array
             ],
             'title' => [
                 'type' => Type::nonNull(Type::string()),
-                'alias' => 'title',
+            ],
+            'likes' => [
+                'type' => Type::listOf(GraphQL::type('Like')),
+                'query' => function (array $args, MorphMany $query) {
+                    return $query->whereRaw('1=1');
```
mfn: This code is never executed => is this expected?

mfn:
> Help wanted

I believe there's single "fix" for this problem: Laravel 5.5 simply generates a different query and I've adapted to test to consider a different assertion 🤷‍♀️

Tests are all green now (ignore StyleCI…)

Is this PR ready for review?

I just noticed _a lot_ of whitespace changes on phpdoc in `SelectFields` and many other kinds of formatting changes => can you please undo this and make "as clean PR as possible"?

tests/Database/SelectFields/InterfaceTests/LikableInterfaceType.php
EdwinDayot: It is now resolved, I added a new test

EdwinDayot:
The PHPDoc has been fixed

tests/Database/SelectFields/InterfaceTests/CommentType.php
```diff
@@ -20,12 +19,7 @@ public function fields(): array
     {
         $interface = GraphQL::type('LikableInterface');
 
-        return [
-                'title' => [
-                    'type' => Type::nonNull(Type::string()),
-                    'alias' => 'title',
-                ],
-            ] + $interface->getFields();
+        return $interface->getFields();
```
mfn: Is this change actually related somehow or just a cleanup?

EdwinDayot: Just some cleanup :)

EdwinDayot:
Readme sounds good to me

mfn:
Preparing the merge now

mfn:
Had to restart the travis jobs, thus it took longer. They were all green the first time, but the feedback loop to github didn't seem to work 🤷‍♀️ Retriggering the jobs on travis fixed it 👍 

Thanks for the PR :)

EdwinDayot:
Nice :)

Issue #481, opened 2019-09-10
Why perform validation method before authorization verification ?


illambo:
### Versions:
- graphql-laravel Version: 2.1
- Laravel Version: 5.8
- PHP Version: 7.1

### Question:
About issue resolved in #407 in my opinion it was better to check first *authorize* method and than perform validation rules, this prevents any superfluous and possibly redundant validation queries.

For example *Rule::exists* check in db if the input data exists, even if I might not have permission to do this. I think this generates bypass the acl controls.

What do you think ? Where I'm wrong ?
Thanks

illambo:
Ping for @rebing and @mfn 
Thanks

rebing:
There are pros and cons with both implementations. If we perform authorization _before_ validation then there might be inputs that should not be there.

Therefore, implementing this method will just replace one problem with another.

illambo:
If I remember correctly in classic laravel http request usually the authorization it is performed before the input validations.
Maybe is it possible implement the "order change" behavior with some configuration option ?

rebing:
Yes, Laravel authorizes before validation. No, we should not make it more complicated by providing a configuration option. 

I'm on the fence about which method to use, so let's wait for @mfn to provide his input. He's on vacay right now 🌴

fvdung:
@rebing Can $context be added as a second argument to the `authorize` function?

rebing:
@fvdung Yes, but it would be a breaking change

mfn:
> If I remember correctly in classic laravel http request usually the authorization it is performed before the input validations.

This is not unambiguously clear.

If we stick to Laravel, what you define in your middleware is the **authentication**, which we can thus say it's always performed before any validation.

Inside your request, you then _usually_ **validate** your request data _and then_ decide if the request is authorized. Because it's not unreasonable to expect to have to use the request data to actually figure out if you are allowed to access something.

In other words: for non trivial tasks, you need to consider the input request to decide if a user is allowed to do something ("authorization"), hence _validation before authorization_ makes sense.

> Can $context be added as a second argument to the authorize function?

This sounds reasonable to me; I was thinking about this with all the refactoring lately when I touched the code surrounding it but haven't gotten around doing it.

> Yes, but it would be a breaking change

Are we sure? I'd love following semver with this package, but adding a new argument which is simply ignored for existing code shouldn't be a problem, or is it? If so, I'd love to get some example code to better understand this, thanks 😄 

### My TL;DR
[...]

PR #487, opened 2019-10-12
Forward PHP engine errors to the application error handler

mfn:
Otherwise PHP errors, like `TypeError` happening (as part of bugs 🤷‍♀️) inside PHP code executed in GraphQL queries are not reported via the application error handler.

Only a regular "internal error" was returned but the actual error was nowhere found in the logs or 3rd party error handling services.

### TODO
- [x] Add test
- [x] Need to find another way to trigger such error because otherwise PHPStan/Larastan complaints because it "sees" that the  code is wrong

### Note
This is a change in behaviour of the library and thus subject to semver requiring a major version bump (as established in https://github.com/rebing/graphql-laravel/issues/481#issuecomment-535380509 ). I.e. the next release has to be `3` after this is merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ CHANGELOG
 
 [Next release](https://github.com/rebing/graphql-laravel/compare/2.1.0...master)
 --------------
+### Changed
+- Forward PHP engine errors to the application error handler
 
 2019-08-27, 2.1.0
 -----------------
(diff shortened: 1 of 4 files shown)


mfn:
FTR, I ran into this in a project I'm working on. There was a bug in the PHP code but because it was an engine error (it isn't in the `Exception` hierarchy but `Error <- Throwable`, I only got back an "internal error" from GraphQL but nothing in the laravel log because `->report()` was not called on it

PR #489, opened 2019-10-12
[BREAKING CHANGE 🚨] Change arguments for `authorize`

mfn:
For background, see https://github.com/rebing/graphql-laravel/issues/481 and specifically https://github.com/rebing/graphql-laravel/issues/481#issuecomment-532926120 and further comments.

The basic change:
- before: `public function authorize(array $args): bool`
- after: `public function authorize($root, array $args, $ctx, ResolveInfo $resolveInfo = null, Closure $getSelectFields = null): bool`

The astute reader will recognize that these are exactly the same arguments passed to `resolve`, and this isn't a coincidence.

This is also made even clearer in the small refactor I did in the code regarding calling them:
```php
            // Authorize
            if (call_user_func_array($authorize, $arguments) != true) {
                throw new AuthorizationError('Unauthorized');
            }

            return call_user_func_array($resolver, $arguments);
```
The only difference is that checking if the returned value and the (automatic) reaction to it (this fact was never different, but due to aligning the arguments it's now even more clear how this works).

Additional changes:
- Don't merge `$arguments[2]` (context) into `$arguments[1]` (args), this seems like a strange thing to do, another one of those implicit behaviours without clear indication as to way (blame didn't offer me much more insights)
- Remove `method_exists` for `getRules`, that method is already there and can never be "not there"

### TODO
- [ ] Add a test making use of all args

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ CHANGELOG
 ### Added
 - Add `wrapType()`, allowing to add more information for queries/mutations [\#496 / albertcito](https://github.com/rebing/graphql-laravel/pull/496)
 ### Changed
+- The signature of `authorize` changed, receiving not the exact same argumenst the resolver would [\#489 / mfn](https://github.com/rebing/graphql-laravel/pull/489)
+  - before: `public function authorize(array $args)`
+  - after: `public function authorize($root, array $args, $ctx, ResolveInfo $resolveInfo = null, Closure $getSelectFields = null): bool`
 - Forward PHP engine errors to the application error handler [\#487 / mfn](https://github.com/rebing/graphql-laravel/pull/487)
 
 2019-08-27, 2.1.0
--- a/Readme.md
+++ b/Readme.md
@@ -590,10 +590,12 @@ An example of Laravel's `'auth'` middleware:
 
 ```php
 use Auth;
+use Closure;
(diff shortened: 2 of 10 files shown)


mfn:
Pinging all involved in https://github.com/rebing/graphql-laravel/issues/481 for feedback @illambo @rebing @fvdung

FTR, I did not go down the suggestion with the trait, I simply couldn't see how this would improve things.

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ CHANGELOG
 
 [Next release](https://github.com/rebing/graphql-laravel/compare/2.1.0...master)
 --------------
+### Changed
+- The signature of `authorized` changed, from only receiving the `$args` to receiving all the same arguments regular `resolve` would also receive [\489 / mfn](https://github.com/rebing/graphql-laravel/pull/489)
```
rebing: Typo, should be `authorize`

mfn:
> How will we version it now - 3.0?

Yes, absolutely has to be!

Actually also for https://github.com/rebing/graphql-laravel/pull/487 => changing the behaviour what kind of exception will be "reported" is also breaking semvar (IMHO / AFAIK)

CHANGELOG.md
mfn: Thanks, fixed!

PR #492, opened 2019-10-15
Allow passing through the rootValue as an option

tuurbo:
As seen in https://github.com/webonyx/graphql-php/blob/master/docs/executing-queries.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ CHANGELOG
 
 [Next release](https://github.com/rebing/graphql-laravel/compare/3.0.0...master)
 --------------
+### Added
+- Allow passing through the `rootValue` as an option [\#492 / tuurbo](https://github.com/rebing/graphql-laravel/pull/492)
 
 2019-10-20, 3.0.0
 -----------------
--- a/src/GraphQL.php
+++ b/src/GraphQL.php
@@ -134,6 +134,7 @@ class GraphQL
         $context = Arr::get($opts, 'context');
         $schemaName = Arr::get($opts, 'schema');
         $operationName = Arr::get($opts, 'operationName');
+        $rootValue = Arr::get($opts, 'rootValue', null);
 
         $schema = $this->schema($schemaName);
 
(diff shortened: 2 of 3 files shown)


tuurbo:
Here is a use case I currently use.

```php
// NotificationMutation.php
$notification = Notification::find($id);
$notification->whatever = 'foobar';
$notification->save();

// send to clients via redis pub/sub
$this->publish('notification', $notification->fresh());
```

```php
function publish($subscriptionName, $rootValue) {
    // check if anyone is subscribing to the $subscriptionName
    $subscription = $this->getSubscription($subscriptionName);

    if ($subscription) {
        // retrieve the schema of the $subscription
        $schema = $this->getSchema($subscription);
    
        // parse $rootValue into a graphql payload
        $payload = \GraphQL::query($schema, [], ['rootValue' => $rootValue]);
        
        // send payload to anyone subscribed to the schema that macthes the $subscriptionName
        Redis::connection('pub')->publish('graphql-channel', json_encode([
            'type' => 'data',
            'id' => $subscription['id'],
            'payload' => $payload,
            'key' => $subscription['key'],
        ]));
    }
}
```

tuurbo:
@mfn Submitted use case and test, Thanks

crissi:
> ```
> publish
> ```

doing graphql subscriptions here? If yes maybe you could care to share som code, would be could to add support for it in the framework

mfn:
Sorry, totally forgot about this PR 🤷‍♀️

I've added a changelog and squashed the commits and released it with https://github.com/rebing/graphql-laravel/releases/tag/3.1.0

mfn:
@tuurbo do you remember why you used `rootValue` and… maybe could you have used context?

I'm in the process of revamping the internal working and wonder if we truly need to support the `rootValue` like this or if `context` would solve this, too?

tuurbo:
@mfn Its been a while since I've looked at this but we still use the code I wrote above today.  I remember copying the apollo graphql JS library and rewriting it into php to suit our needs. https://github.com/apollographql/graphql-subscriptions/blob/v0.1.5/src/pubsub.ts#L155-L156 Not sure if the recommended way it's done now is any different.

Passing the data as context seems weird to me but I suppose it would work. I don't know enough about the inner workings of graphql to say for sure.

mfn:
@tuurbo thanks for your quick response, I found out I don't need to remove support for it

Issue #491, opened 2019-10-15
A field implemented through Interface does not use the proper alias


EdwinDayot:
### Versions:
- graphql-laravel Version: 2.1.1

### Description:

If a `Type` is implementing an `Interface` defining a field with an alias, and we try to overwrite it in the `Type`, the alias of the `Interface` will always be used.

### Steps To Reproduce:

```php
class InterfaceExample {
  public function fields() {
    return [
      'a_field' => [
        'alias' => 'an_alias' // This will be used
      ]
    ];
  }
}

class TypeExample {
  public function fields() {
    $fields = GraphQL::type('InterfaceExample');
    return $fields->getFields() + [
      'a_field' => [
        'alias' => 'another_alias' // Instead of this
      ]
    ];
  }
}
```

EdwinDayot:
Interfaces are quite complex to handle, I think we should move everything handling Interfaces into Closures executed as needed so we have access to the _real_ type.

mfn:
Would be great if you could add a test for the existing behaviour to master

EdwinDayot:
Sure, I will

EdwinDayot:
This bug has been reintroduced with version 5.0.0

mfn:
💥 

How? Not pretending to understand where exactly :) thought things were covered in tests.

Can you check this and make a fix?

illambo:
Strange, i remember the test [_testGeneratedInterfaceFieldInlineFragmentsAndAlias_ ](https://github.com/rebing/graphql-laravel/pull/607/commits/c0d12f709e211ef0238e5d9a688bac096295a653) covered the alias case.
Maybe I did something wrong ?

illambo:
Mmm, yes cover the alias case but not exactly the current case.
I try to verify better with a specific unit test, sorry for that.

illambo:
Ok, I try in local with specific test and seems works properly, I think the problem is in the use of "+" for array merge (better user array_merge), pls try with this: 

```php
class TypeExample {
  public function fields() {
    $interface = GraphQL::type('InterfaceExample');
    return array_merge($interface->getFields(), [
      'a_field' => [
        'alias' => 'another_alias' // Instead of this
      ]
    ]);
  }
}
```

EdwinDayot:
I'll try this ! Thanks !

PR #494, opened 2019-10-16
Added test for alias of field in interface implemented types

EdwinDayot:
#491


mfn:
Btw, when clicking on the SyleCI "Details", you can also download the diff to apply to fix the violations.

I see that 5.5 tests break. Well, that's a bummer. I didn't inspect it yet, but I guess it's related to the builder generating a slightly different SQL (which would kind of negative the SQL tests that way …)

Issue #497, opened 2019-10-20
Should the project switch to a different code / style / enforcer


mfn:
Spin-off from https://github.com/rebing/graphql-laravel/pull/485#issuecomment-544126161

When I started to increase my contributions to the project, my first and biggest gripe was that there was not coherent code style in the code base.

Having had experience with https://styleci.io/ , I simple went ahead to "fix it quickly" and ensured that all source files where on the same level. An additional benefit of this integration was that, independent on who contributes, the style is implicitly enforced as a dedicated check which needs to be green before a merge can happen; ensuring the code base style stays coherent..

From the referenced issue:
> Do you have some command to verify the style code in my local environment? Something like that phpcs --standard=phpcs.xml --extensions=php src.

My answer:
> Unfortunately no, StyleCI entirely works "in the cloud"; so only after a push you get back the information via Github. The "best" you can do then, instead of manually fixing it, is to simply download the diff and apply it, push again (or, as I usually do it, rebase)

### TL;DR: I acknowledge this is not ideal and I'm open for changing this

Things to be considered for an alternative:
- [x] preferable open source, as is this package 😄 
- [x] easy to run locally for everyone
- [ ] be able to automatically commit the fixes via a CI pipeline (github actions?)

The astute reader will notice the above points don't talk about _the actual code style_. For a start, no one specifically complained about it. And personally I don't care _too much_ about which actual style, as long as some criteria can be met:
- should be a well acknowledged / used style in the OSS community
  Currently it's using the "Laravel" style guide as implemented by StyleCI; it being simply a Laravel package. Personally though I've no feelings that it has to be "the Laravel" style
- should be enforceable within the IDE automatically "to a high degree"
  I'm fine that an external / OSS tool can be used to always align the source to the "final form", but an IDE should be able to "come close" to it so ideally the necessary changes aren't too much every time
  My experience with the current setup and e.g. PhpStorm is quite a good one. In 99% of the cases of StyleCI it's usually only some `use` statement reordering when I work with it

Feedback / actionables very much welcome. Personally outside StyleCI I've not much experience so basically this issue relies on someone fill the void a bit here :)

crissi:
using phpfixer with @PSR2 rule set would be nice supplemented with what ever custom rule makes sense.

But most importantly the possiblity to run it locally.

Another alternative is for the pipeline to fix the styling and commit on push.

mfn:
[...]

PR #502, opened 2019-10-24
Switch from StyleCI to PHP-CS fixer

crissi:
now developers can run composer fix to fix styling issues and more easily set up their IDE because we are using phpcsfixer

and travis will fail and show the error diff from php-cs-fixer

## TODOs
- [ ] Don't know how to disable styleci
- [x] I am running the linter on every environment in travis, this could probably be changed for faster a pipeline. Don't know how easy that is!?

### Links
- Connected to https://github.com/rebing/graphql-laravel/issues/497


mfn:
> I am running the linter on every environment in travis, this could probably be changed for faster a pipeline. Don't know how easy that is!?

Done, pushed a commit! I just use env variables for this and we just need to make sure that at _least for one_ job it's set. Same approach used for e.g. phpstan.

I also fixed the failing lumen integration test. I boy I hate them, but it's better then nothing. They rely on scripting the manual steps necessary and easily fail due to the sensitivity of the regex when the source file changes even formatting (which happened in this case for `config/config.php`). Anyway, all green!

I also took the liberty to simply remove the extended tests from 5.8; it's EOL anyway and only wil receive security fixes which shouldn't be critical for the extended test.

> Don't know how to disable styleci

Only admins can do that (which I'm not).

But before proceeding I would like to do some testing/feedbacking

.php_cs.dist
```diff
@@ -0,0 +1,55 @@
+<?php
+
+/**
+ * Rules we follow are from PSR-2 as well as the rectified PSR-2 guide.
```
mfn: Is there a blueprint where you got this whole fle from?

.php_cs.dist
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+$finder = PhpCsFixer\Finder::create()
+    ->exclude('node_modules')
+    ->exclude('vendor')
+    ->in(__DIR__);
+
+return PhpCsFixer\Config::create()
+    ->setRules([
+        '@PSR2' => true,
```
mfn: So. I was wondering since it said `@PSR2` at the top, why the other rules?

I removed them all and ran `fix` and it did quite some changes, like these:
```diff
diff --git a/src/GraphQL.php b/src/GraphQL.php
index c5d3b0d..9562554 100644
--- a/src/GraphQL.php
+++ b/src/GraphQL.php
@@ -239,10 +239,12 @@ class GraphQL

         if (! $type instanceof TypeConvertible) {
             throw new TypeNotFound(
-                sprintf('Unable to convert %s to a GraphQL type, please add/implement the interface %s',
+                sprintf(
+                    'Unable to convert %s to a GraphQL type, please add/implement the interface %s',
                     get_class($type),
                     TypeConvertible::class
-                ));
+                )
+            );
         }

         foreach ($opts as $key => $value) {
diff --git a/src/Support/ResolveInfoFieldsAndArguments.php b/src/Support/ResolveInfoFieldsAndArguments.php
index 34aceea..bed6194 100644
--- a/src/Support/ResolveInfoFieldsAndArguments.php
+++ b/src/Support/ResolveInfoFieldsAndArguments.php
@@ -130,12 +130,16 @@ class ResolveInfoFieldsAndArguments
                 $spreadName = $selectionNode->name->value;
                 if (isset($this->info->fragments[$spreadName])) {
                     $fragment = $this->info->fragments[$spreadName];
-                    $fields = array_merge_recursive($this->foldSelectionSet($fragment->selectionSet, $descend),
-                        $fields);
+                    $fields = array_merge_recursive(
+                        $this->foldSelectionSet($fragment->selectionSet, $descend),
+                        $fields
+                    );
                 }
             } elseif ($selectionNode instanceof InlineFragmentNode) {
-                $fields = array_merge_recursive($this->foldSelectionSet($selectionNode->selectionSet, $descend),
-                    $fields);
+                $fields = array_merge_recursive(
+                    $this->foldSelectionSet($selectionNode->selectionSet, $descend),
+                    $fields
+                );
             }
         }
```
or
```diff
diff --git a/tests/Database/MutationValidationUniqueWithCustomRulesTests/MutationValidationUniqueWithCustomRulesTest.php b/tests/Database/MutationValidationUniqueWithCustomRulesTests/MutationValidationUniqueWithCustomRulesTest.php
index 82a2932..e1a7581 100644
--- a/tests/Database/MutationValidationUniqueWithCustomRulesTests/MutationValidationUniqueWithCustomRulesTest.php
+++ b/tests/Database/MutationValidationUniqueWithCustomRulesTests/MutationValidationUniqueWithCustomRulesTest.php
@@ -36,7 +36,8 @@ GRAPHQL;
             ],
         ]);

-        $this->assertSqlQueries(<<<'SQL'
+        $this->assertSqlQueries(
+            <<<'SQL'
```
Basically it felt more aggressive.

I liked that because for example, in the _current_ version these two patterns are allowed:
```php
        $this->assertSqlQueries(
            <<<'SQL'
```
and
```php
        $this->assertSqlQueries(<<<'SQL'
```
Nether one is touched and converted to the other.

Yet when I remove all the settings, it's converted the former form always.

I think this is better, the source will be more unified and leaves less room for interpretation.

.php_cs.dist
```diff
@@ -0,0 +1,55 @@
...
+        'array_syntax' => [ 'syntax' => 'short' ],
```
mfn: Immediately noticed though, per my above comment, removing this one here means it will not enforce `$foo = array();` being converted to `$foo = [];`

So I guess just `@PSR2` isn't specific enough and still has room for interpretation.

.php_cs.dist
crissi: https://gist.github.com/codfish/c77d348820c1c6b4ebe4a66dc2291c74#file-php_cs-dist

.php_cs.dist
crissi: yeah @PSR2 is not enough to get the "prettiest" formatting:-)

src/Support/Field.php
```diff
@@ -259,7 +259,7 @@ public function __get($key)
     {
         $attributes = $this->getAttributes();
 
-        return isset($attributes[$key]) ? $attributes[$key] : null;
+        return $attributes[$key] ?? null;
```
mfn: Now this is really nice

src/Support/SelectFields.php
```diff
@@ -102,16 +102,16 @@ public static function getSelectableFieldsAndRelations(
...
-            return function ($query) use ($with, $select, $customQuery, $requestedFields, $ctx) {
-                if ($customQuery) {
-                    $query = $customQuery($requestedFields['args'], $query, $ctx);
-                }
-
-                $query->select($select);
-                $query->with($with);
-            };
         }
+
```
mfn: Oh my 🤔

crissi:
Maybe this? https://github.com/matt-allan/laravel-code-style

Sendt: Tuesday, November 5, 2019 10:26:52 PM
Emne: Re: [rebing/graphql-laravel] [WIP] add php cs fixer (#502)


@mfn commented on this pull request.

I've found https://gist.github.com/laravel-shift/cab527923ed2a109dda047b97d53c200 and applied and pushed the changes for review.

Not sold yet on the config though :)

I wish there was a better way then copypasting some gist around; maybe there is a package with a maintained "Laravel" compatible PHP-CS fixer config we could just be pulled it as a dev dependency?

In src/Support/Field.php<https://github.com/rebing/graphql-laravel/pull/502#discussion_r342804866>:

> @@ -259,7 +259,7 @@ public function __get($key)

     {

         $attributes = $this->getAttributes();



-        return isset($attributes[$key]) ? $attributes[$key] : null;

+        return $attributes[$key] ?? null;



Now this is really nice

In src/Support/SelectFields.php<https://github.com/rebing/graphql-laravel/pull/502#discussion_r342805156>:

>          }

+



Oh my 🤔

mfn:
I pushed some updates:
- we need to use matt-allan/laravel-code-style dev-master because alphabetical sorting of imports is fixed there but it's not yet released
  - and since we already use alphabetical imports, it wouldn't make sense to go back and forth on this
- the package causes conflicts with older PHP versions we test (7.1) and older Laravel rrameworks (5.5). But since we only need this be running for _one_ job at least, I simply made it to be _uninstalled_ for all parts we don't run the linter

mfn:
Also added a PR template which mentions how to fix the code style

EdwinDayot:
+1 for this ! Being able to fix everything locally is awesome. It would be great being able to do the same thing for travis, but I can imagine what a mess it could be.

johannesnagl:
I've added a small fix in here: https://github.com/crissi/graphql-laravel/pull/9

mfn:
@johannesnagl 
> I've added a small fix in here: [crissi#9](https://github.com/crissi/graphql-laravel/pull/9)

Thanks, already cherry-picked plus some other fixes!

crissi:
So I gues we just need to let it pass the styleci check and then after this PR is merged the styleci is gone?!?!

mfn:
@rebing can you also please remove StyleCI? Thanks 🤗

PR #503, opened 2019-10-25
add morph type to returned models

crissi:
when you retrieve the morphed models in your query resolver the type is missing so don't know what type it is

**Are we going to continue to support 5.5? can see the tests are failing for that version**

### DIFF
--- a/src/Support/SelectFields.php
+++ b/src/Support/SelectFields.php
@@ -27,7 +27,7 @@ class SelectFields
     /** @var array */
     private $relations = [];
 
-    const FOREIGN_KEY = 'foreignKey';
+    const ALWAYS_RELATION_KEY = 'ALWAYS_RELATION_KEY';
 
     /**
      * @param  ResolveInfo  $info
@@ -142,7 +142,7 @@ class SelectFields
             }
 
             // Always select foreign key
-            if ($field === self::FOREIGN_KEY) {
+            if ($field === self::ALWAYS_RELATION_KEY) {
                 self::addFieldToSelect($key, $select, $parentTable, false);
                 continue;
             }
@@ -212,7 +212,11 @@ class SelectFields
                             $segments = explode('.', $foreignKey);
                             $foreignKey = end($segments);
                             if (! array_key_exists($foreignKey, $field)) {
-                                $field['fields'][$foreignKey] = self::FOREIGN_KEY;
+                                $field['fields'][$foreignKey] = self::ALWAYS_RELATION_KEY;
+                            }
+
+                            if (is_a($relation, MorphMany::class) || is_a($relation, MorphOne::class)) {
+                                $field['fields'][$relation->getMorphType()] = self::ALWAYS_RELATION_KEY;
                             }
                         }
 
@@ -295,7 +299,8 @@ class SelectFields
 
                 default:
                     throw new RuntimeException(
-                        sprintf("Unsupported use of 'privacy' configuration on field '%s'.",
+                        sprintf(
+                            "Unsupported use of 'privacy' configuration on field '%s'.",
                             $fieldObject->name
                         )
                     );
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 --------------
 ### Fixed
 - Fix validation rules for non-null list of non-null objects [\#511 / crissi](https://github.com/rebing/graphql-laravel/pull/511/files)
+- Add morph type to returned models [\#503 / crissi](https://github.com/rebing/graphql-laravel/pull/503)
 
 2019-10-23, 3.1.0
 -----------------
--- a/tests/Database/SelectFields/MorphRelationshipTests/CommentType.php
+++ b/tests/Database/SelectFields/MorphRelationshipTests/CommentType.php
@@ -1,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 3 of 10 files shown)


mfn:
> Are we going to continue to support 5.5? can see the tests are failing for that version

Thanks for asking, I haven't yet disclosed my thoughts on this.

Basically I'm following https://laravel.com/docs/master/releases
image

That's why I also removed tests for 5.6 and 5.7 without much discussion via https://github.com/rebing/graphql-laravel/pull/488

I basically took the easy way out: I just stick to the LTS table. Once a version is beyond security fixes _it can_ be removed on first sign of troubles.

crissi:
> > Are we going to continue to support 5.5? can see the tests are failing for that version
> 
> Thanks for asking, I haven't yet disclosed my thoughts on this.
> 
> Basically I'm following https://laravel.com/docs/master/releases
> image
> 
> That's why I also removed tests for 5.6 and 5.7 without much discussion via #488
> 
> I basically took the easy way out: I just stick to the LTS table. Once a version is beyond security fixes _it can_ be removed on first sign of troubles.

Ok will make a fix for laravel 5.5

crissi:
how do I actually run this for Laravel 5.5? I tried removing all other versions from illuminate/contracts, illuminate/support in the  composer.json, but gets alot of composer errors.

mfn:
I believe from `.travis.yml`, these lines from the `install` section are relevant:
```
composer require "illuminate/support:${LARAVEL}" --no-interaction --no-update
composer install --prefer-dist --no-interaction --no-suggest
```

crissi:
Thanks, will try

Sendt: Monday, October 28, 2019 12:24:53 PM
Emne: Re: [rebing/graphql-laravel] add morph type to returned models (#503)


I believe from .travis.yml, these lines from the install section are relevant:

composer require "illuminate/support:${LARAVEL}" --no-interaction --no-update
composer install --prefer-dist --no-interaction --no-suggest

src/Support/SelectFields.php
```diff
@@ -214,6 +220,10 @@ protected static function handleFields(
                             if (! array_key_exists($foreignKey, $field)) {
                                 $field['fields'][$foreignKey] = self::FOREIGN_KEY;
                             }
+
+                            if (method_exists($relation, 'getMorphType')) {
```
mfn: Shouldn't this rather use `is_a` like in the if condition a few lines above (for the two morph types entering this condition)?

src/Support/SelectFields.php
```diff
@@ -147,6 +148,11 @@ protected static function handleFields(
                 continue;
             }
 
+            if ($field === self::MORPH_TYPE_KEY) {
```
mfn: How about merging with the if condition before, as in `if ($field === … || $field === …)`. Seems to be both actions are the same, anyway?

mfn: Which begs the question: do we need the separate constant `MORPH_TYPE_KEY` really? It behaves exactly the same as the other constant.

Mayne we should change the constant to better express the intent, something like: `ALWAYS_RELATION_KEY` or something.

What do you think?

tests/Database/SelectFields/RelationshipsTest/MorphRelationshipTest.php
```diff
@@ -0,0 +1,121 @@
...
+
+        $result = $this->graphql($query, [
+            'expectErrors' => true,
+        ]);
+
+        $this->assertSqlQueries(
+            <<<'SQL'
+select "users"."id" from "users";
+select "posts"."id", "posts"."title", "posts"."user_id" from "posts" where "posts"."user_id" in (?, ?) order by "posts"."id" asc;
+select "likes"."id", "likes"."user_id", "likes"."likable_id", "likes"."likable_type" from "likes" where "likes"."likable_id" in (?) and "likes"."likable_type" = ?;
```
mfn: I feel stupid to ask, but when I leave the tests but rollback the changes in `SelectFields`, this test here only fails due to the different SQL statement (`"likes"."likable_type"` is not included).

But the result assertions are exactly the same. So from the outside, it's not observable.

Can you elaborate?

tests/Database/SelectFields/RelationshipsTest/MorphRelationshipTest.php
```diff
@@ -0,0 +1,121 @@
...
+                    ],
+                    [
+                        'id' => $otherUser->id,
+                        'posts' => [],
+                    ],
+                ],
+            ],
+        ];
+        $this->assertEquals($expectedResult, $result);
+        $this->assertEquals($expectedResult, $result);
```
mfn: Redundant

src/Support/SelectFields.php
crissi: good idea

tests/Database/SelectFields/RelationshipsTest/MorphRelationshipTest.php
crissi: I refactored the test to add test assertions inside the resolve method, to better reflect what it is fixing

mfn: Oh wow, totally missed that.

But ... it's also _not_ failing for the master branch version of `SelectFields`. What am I missing?

crissi: I forgot to remove the expectError thingy

PR #508, opened 2019-10-26
fixes re-declare issue

Artem-Schander:
```
ErrorException: Cannot declare class Illuminate\Routing\Controller, because the name is already in use
/var/www/html/vendor/rebing/graphql-laravel/src/GraphQLLumenServiceProvider.php:24
/var/www/html/vendor/laravel/lumen-framework/src/Application.php:193
/var/www/html/app/Providers/AppServiceProvider.php:35
```

### DIFF
--- a/src/GraphQLLumenServiceProvider.php
+++ b/src/GraphQLLumenServiceProvider.php
@@ -20,7 +20,7 @@ class GraphQLLumenServiceProvider extends GraphQLServiceProvider
 
     public function register()
     {
-        if (class_exists('Laravel\Lumen\Routing\Controller')) {
+        if (class_exists('Laravel\Lumen\Routing\Controller') && ! class_exists('Illuminate\Routing\Controller')) {
             class_alias('Laravel\Lumen\Routing\Controller', 'Illuminate\Routing\Controller');
         }
 


mfn:
PR seems legit to me.

Yet I wonder: what are the actual circumstances? Is this a recent regression?

We've a very basic Lumen integration test and I don't remember seeing this.

Thanks

Artem-Schander:
Hi Markus,

i have no idea what the cause is. Currently I'm migrating a huge project from folklore to this implementation (BTW many thanks) and the exception you see above is the only issue so far. So I thought I'll do a PR.
I could try to find some time and create a minimalistic example project to reproduce this issue if this is important to you.

For now I have a custom ServiceProvider, but it would be nice if you could merge this one.

mfn:
To you have the package `illuminate/routing` in your Lumen `composer.json` as root dependency?

If not as root, may as any kind of dependency? `composer depend illuminate/routing` would show this.

Artem-Schander:
Does not seem like that
```

**Comment by mfn** [comment]
Please post your `composer.json`

**Comment by Artem-Schander** [comment]
```
{
    "name": "laravel/lumen",
    "description": "The Laravel Lumen Framework.",
    "keywords": ["framework", "laravel", "lumen"],
    "license": "MIT",
    "type": "project",
    "require": {
        "php": ">=7.0",
        "laravel/lumen-framework": "5.5.*",
        "vlucas/phpdotenv": "~2.2",
        "illuminate/notifications": "^5.5",
        "illuminate/mail": "^5.5",
        "tymon/jwt-auth": "^1.0",
        "barryvdh/laravel-cors": "^0.10.0",
        "intervention/image": "^2.4",
        "league/csv": "^9.1",
        "graham-campbell/flysystem": "^4.0",
        "respect/validation": "^1.1",
        "cocur/slugify": "^3.0",
        "moneyphp/money": "^3.0",
        "php-http/message": "^1.6",
        "php-http/guzzle6-adapter": "^1.1",
        "florianv/swap": "^3.2",
        "irazasyed/larasupport": "^1.6",
        "tecnickcom/tcpdf": "^6.2",
        "illuminate/redis": "^5.5",
        "predis/predis": "^1.1",
        "ignited/laravel-omnipay": "2.*",
        "omnipay/paypal": "^2.6",
        "omnipay/stripe": "~2.0",
        "kalnoy/nestedset": "^4.3",
        "sabre/xml": "^2.1",
        "rebing/graphql-laravel": "^3.1"
    },
    "require-dev": {
        "fzaninotto/faker": "~1.4",
        "mockery/mockery": "~1.0",
        "phpunit/phpunit": "~6.0",
        "tightenco/mailthief": "^0.3.14"
    },
    "autoload": {
        "psr-4": {
            "App\\": "app/"
        },
        "files": [
            "app/helpers.php"
        ]
    },
    "autoload-dev": {
        "classmap": [
            "tests/",
            "database/"
        ]
    },
    "scripts": {
        "post-root-package-install": [
            "php -r \"copy('.env.example', '.env');\""
        ]
    },
    "minimum-stability": "dev",
    "prefer-stable": true,
    "config": {
        "optimize-autoloader": true
    }
}
```

**Comment by mfn** [comment]
Based on this composer configuration, I installed all this dependencies and searched for the controller file:
```
$ find . -type f -name Controller\.php
./vendor/laravel/lumen-framework/src/Routing/Controller.php
```
I.e. there is no other file controller file exept the one from lumen itse.f

Something is not matching here.

mfn:
That was my "Plan B" question. Only you can figure this out.

Something like `dump(debug_backtrace());` or `echo (new \RuntimeException())->getTraceAsString(), "\n";` right in that method should give you an idea if it's called twice.

To be clear:
I'm not against the PR, but I don't like merging stuff I don't understand or can't reproduce.

On paper, the PR looks innocent and fine.

But if you think about it, how this can be possible without the actual Laravel routing component being there, then it starts to get fishy. And the one thing Lumen does on its own _is_ routing so mixing Laravel routing component into Lumen is weird; like "why not using Laravel then"-weird.

Thanks for understanding

mfn:
@Artem-Schander have you made any progress debugging this?

Artem-Schander:
no, not yet.
I'll try to get my hands on it on the weekend.

mfn:
Whilst the change looks trivial, I think what's missing here is step 0: understanding _what the actual cause_ is. Until there is progress in this area, I'll hold off merging this for now.

Thanks for your contribution!

mfn:
@Artem-Schander sorry for all the confusion, after @ulrichdahl made https://github.com/rebing/graphql-laravel/pull/664 we had some discussion about this topic again and to me most revealing was the description from https://github.com/rebing/graphql-laravel/pull/664/commits/e854d83d8e9552307b0cbe6044bebee3a431dc5a
> It is rather simple really. 
> The default testcase in lumen does an app refresh before each test in the setUp method, which does a bootstrap.
> This means in the same process all classes are being re autoloaded, which is not a problem since it is done with require_once. But the GraphQL does not fall under that since is being re executed and not required. 
>
> So there is an unmet requirement of the register function, it needs to be process idempotent.

I realized this approach was fine, so to honor both your work I:
- reopened and merged this PR
- edited ❗ your original commit to add above exact text and references @ulrichdahl

Sorry for the long delay and thanks for everyones contribution and patience!

----
Note: this is only merged into master now which builds on the next major release of graphql-php bringing breaking changes, so this will only go into the next major release I hope to make in September (~2-3 weeks from now).

Issue #505, opened 2019-10-25
Trim mutation arguments like laravel TrimStrings middleware


tpetry:
Currently it is problematic to apply validation rules to the input of mutations: For Laravel http requests all spaces at the beginning and end of input values are removed using the `TrimStrings` middleware. So Validation Rules like `between:10,50` can work on the real length without whitespaces.

However, this is not possible with GraphQL queries so far. The validations work directly on the data from the query and thus on strings with spaces. And there is no way to change it or hook into it to change the behaviour.

It would be very nice if there was an option to apply the trimming of whitespaces recursively to the arguments of GraphQL queries.

mfn:
Hmm, there was another issue a few months ago => https://github.com/rebing/graphql-laravel/issues/248

It's true the library doesn't provide anything out of the box. The "graphql way" would be to use directives, which aren't supported though.

But I think there could be an easy workaround with your own middlware:
```php
    public function handle(Request $request, Closure $next)
    {
        $input = $request->input();
        // trim…
        $request->merge($input);

        return $next($request);
    }
```
For this to work, you would not pass the data as `variables` from the client though, not within the query. Then you can recursively trim it.

tpetry:
> For this to work, you would not pass the data as variables from the client though, not within the query. Then you can recursively trim it.
Yeah and that's the problem. This approach may work if you are the only one using the api. But if it's an public api you can't enforce this rules.

I thought maybe a small snippet somewhere could be implemented to trim the values before they are used for everything.

mfn:
Certainly, but I prefer there to be used an "official channel" to accomplish this instead of wiping up something custom.

tpetry:
What do you mean by official? Facebook only publishes the graphql spec and reference implementation which webonyx ported to php. Logic like trimming whitespaces has done by an upper layer (like laravel implementing ontop of basic php http request).
So imho your library is the correct place to implement this functionality in the layers.

mfn:
Apologies for being unclear. Basically what I said in https://github.com/rebing/graphql-laravel/issues/505#issuecomment-546481586 but more elaborate:

Yes, I think implementing this as part of the library is the appropriate. That's why I also didn't close this issue.

With "official" I meant: I don't want to introduce some artificial trim layer "because reasons".

Within the GraphQL spec, there's a clear candidate for implementing this: [directives](https://graphql.github.io/graphql-spec/June2018/#sec-Language.Directives).

[...]

PR #517, opened 2019-11-04
Alias for arguments

crissi:
especially useful for mutation where you would like the inputs to match the database columns

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ CHANGELOG
 --------------
 ### Added
 - Allow passing through an instance of a `Field` [\#521 / georgeboot](https://github.com/rebing/graphql-laravel/pull/521/files)
-- add the ability to alias query and mutations arguments as well as input objects [\#517 / crissi](https://github.com/rebing/graphql-laravel/pull/517/files)
+- Add the ability to alias query and mutations arguments as well as input objects [\#517 / crissi](https://github.com/rebing/graphql-laravel/pull/517/files)
 ### Fixed
 - Fix validation rules for non-null list of non-null objects [\#511 / crissi](https://github.com/rebing/graphql-laravel/pull/511/files)
 - Add morph type to returned models [\#503 / crissi](https://github.com/rebing/graphql-laravel/pull/503)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,15 @@ CHANGELOG
 
 [Next release](https://github.com/rebing/graphql-laravel/compare/3.1.0...master)
 --------------
+### Added
+- Allow passing through an instance of a `Field` [\#521 / georgeboot](https://github.com/rebing/graphql-laravel/pull/521/files)
+- add the ability to alias query and mutations arguments as well as input objects [\#517 / crissi](https://github.com/rebing/graphql-laravel/pull/517/files)
 ### Fixed
 - Fix validation rules for non-null list of non-null objects [\#511 / crissi](https://github.com/rebing/graphql-laravel/pull/511/files)
+- Add morph type to returned models [\#503 / crissi](https://github.com/rebing/graphql-laravel/pull/503)
+### Changed
+- Switch Code Style handling from StyleCI to PHP-CS Fixer [\#502 / crissi](https://github.com/rebing/graphql-laravel/pull/502)
+- More control over validation through optional user-generated validator by introducing `getValidator()` [\#531 / mailspice](https://github.com/rebing/graphql-laravel/pull/531)
 
 2019-10-23, 3.1.0
 -----------------
--- a/Readme.md
+++ b/Readme.md
@@ -3,11 +3,10 @@
 [![Latest Stable Version](https://poser.pugx.org/rebing/graphql-laravel/v/stable)](https://packagist.org/packages/rebing/graphql-laravel)
 [![codecov](https://codecov.io/gh/rebing/graphql-laravel/branch/master/graph/badge.svg)](https://codecov.io/gh/rebing/graphql-laravel)
 [![Build Status](https://travis-ci.org/rebing/graphql-laravel.svg?branch=master)](https://travis-ci.org/rebing/graphql-laravel)
-[![Style CI](https://styleci.io/repos/68595316/shield)](https://styleci.io/repos/68595316)
 [![License](https://poser.pugx.org/rebing/graphql-laravel/license)](https://packagist.org/packages/rebing/graphql-laravel)
 [![Get on Slack](https://img.shields.io/badge/slack-join-orange.svg)](https://join.slack.com/t/rebing-graphql/shared_invite/enQtNTE5NjQzNDI5MzQ4LWVjMTMxNzIyZjBlNTFhZGQ5MDVjZDAwZDNjODA3ODE2NjdiOGJkMjMwMTZkZmNhZjhiYTE1MjEyNDk0MWJmMzk)
 
-### Note: these are the docs for 2.*, [please see the `v1` branch for the 1.* docs](https://github.com/rebing/graphql-laravel/tree/v1#laravel-graphql)
+### Note: these are the docs for the current release, [please see the `v1` branch for the 1.* docs](https://github.com/rebing/graphql-laravel/tree/v1#laravel-graphql)
 
 Uses Facebook GraphQL with Laravel 5.5+. It is based on the PHP implementation [here](https://github.com/webonyx/graphql-php). You can find more information about GraphQL in the [GraphQL Introduction](http://facebook.github.io/react/blog/2015/05/01/graphql-introduction.html) on the [React](http://facebook.github.io/react) blog or you can read the [GraphQL specifications](https://facebook.github.io/graphql/). This is a work in progress.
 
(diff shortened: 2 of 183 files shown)


mfn:
I'm not against this feature.

But I don't like that it's _forced_ upon everyone.

We support the concept of `'alias'` on fields but this is optional: only if you make use of `$getSelectFields()` you opt-in to that processing and the overhead of it.

Not so with this approach as it's hardcoded in the resolver (and I would find it a bit extreme to override the resolver just to be able to disable this).

It feels this is in a similar space as the idea of https://github.com/rebing/graphql-laravel/issues/505 , namely to activate some optional kind of processing in various stages here. I think this belongs in a bigger picture (which I don't have at this moment).

crissi:
> I'm not against this feature.
> 
> But I don't like that it's _forced_ upon everyone.
> 
> We support the concept of `'alias'` on fields but this is optional: only if you make use of `$getSelectFields()` you opt-in to that processing and the overhead of it.
> 
> Not so with this approach as it's hardcoded in the resolver (and I would find it a bit extreme to override the resolver just to be able to disable this).
> 
> It feels this is in a similar space as the idea of #505 , namely to activate some optional kind of processing in various stages here. I think this belongs in a bigger picture (which I don't have at this moment).

I see you point, but I been using this for quite a time right now and I think it such a common usecase, especially for mutations.

But it could also be implemented as as middleware of some sort.

mfn:
> But it could also be implemented as as middleware of some sort.

I was thinking in that direction but when I look at the PR, if I understand it correctly, the changes have to be done _after_ validation but _before_ invoking the resolver (actually not even sure why/how it works correctly with the $getSelectFields closure).

So this would require a *very* specific _kind of_ middleware, like "RightInTheMiddleOfGetResolverMiddleware" 😜 or something.

This middleware thing would have to receive and return all arguments but should probably just receive everything, maybe even `$this->type()`.

> I been using this for quite a time right now and I think it such a common usecase, especially for mutations.

How did you go about it? Did you override `\Rebing\GraphQL\Support\Field::getResolver`?

crissi:
> > But it could also be implemented as as middleware of some sort.
> 
> I was thinking in that direction but when I look at the PR, if I understand it correctly, the changes have to be done _after_ validation but _before_ invoking the resolver (actually not even sure why/how it works correctly with the $getSelectFields closure).
> 
That one still receives the original input
> So this would require a _very_ specific _kind of_ middleware, like "RightInTheMiddleOfGetResolverMiddleware" 😜 or something.

:grin:

> 
> This middleware thing would have to receive and return all arguments but should probably just receive everything, maybe even `$this->type()`.
> 
> > I been using this for quite a time right now and I think it such a common usecase, especially for mutations.
> 
> How did you go about it? Did you override `\Rebing\GraphQL\Support\Field::getResolver`?

yes

mfn:
> > How did you go about it? Did you override `\Rebing\GraphQL\Support\Field::getResolver`?
> 
> yes

Ok, what I thought / feared 😄  But it highlights the shortcomings, thanks.

> That one still receives the original input

But I then think this isn't quite appropriate, per you initial description:
> where you would like the inputs to match the database columns

This would mean the actually `SelectFields` would **not** benefit from that.


----

Maybe it needs some more hooks into the system or, as I stated already, looking into GraphQL directives. But I don't want to incur overhead for a feature, as useful as it sounds, which may not be used at all by consumers of the library (aka opt-in).

My TL;DR is that I don't think modifying the resolver this way is the right approach and as such I don't want to accept the PR (at least in its current stage).

crissi:
> > > How did you go about it? Did you override `\Rebing\GraphQL\Support\Field::getResolver`?
> > 
> > 
> > yes
> 
> Ok, what I thought / feared 😄 But it highlights the shortcomings, thanks.
> 
> > That one still receives the original input
> 
> But I then think this isn't quite appropriate, per you initial description:
> 
> > where you would like the inputs to match the database columns
> 
> This would mean the actually `SelectFields` would **not** benefit from that.
> 

I guess this could be added here to, I just never actual thought about that since I do not use the https://github.com/rebing/graphql-laravel#type-relationship-query very often

> Maybe it needs some more hooks into the system or, as I stated already, looking into GraphQL directives. But I don't want to incur overhead for a feature, as useful as it sounds, which may not be used at all by consumers of the library (aka opt-in).
> 
> My TL;DR is that I don't think modifying the resolver this way is the right approach and as such I don't want to accept the PR (at least in its current stage).

fair

crissi:
Should I close this or would you accept this with the modifications to the selectFields?

of course an option that enables this could also be added?

mfn:
Thanks for asking, I didn't close it as I didn't want to signal an outright rejection, just the implementation.

Whatever fits you, re-iterating this PR or making a new one, I don't mind either.

crissi:
but you want one that does not modify the getResolver directly?

mfn:
Not per se, I try to formulate what I think we should be aware of:
- keep the API simple
- don't force it on users if not needed

I already think it's not really elegant that `resolve` receives a `$getSelectFields` argument and which there was another way, but here we already tried to remove hardcoded "opt-in". Before that change everyone was always getting a `SelectFields` class even if you didn't use it.

OTOH when I think about it, a feature like supporting aliases for input argument might not even something that should be easily turned off / on. It can be useful regardless of the database context.

But when I look at the current implemention `AliasArguments`, it already has to do quite some work and that is even only half of the story, because parts are in the external dependency.

georgeboot:
I would also love to see a feature like this.
My database columns are all `snake_case`, whereas the fields in my types are all `camelCase`.

I now find my self constantly mapping them by hand, whereas a `new Model($args['input'])` in combination with `$fillable` would be much easier.

Regarding the added overhead: have you done any research on how big the overhead is?
In my opinion, If te API remains simple, I don't think we should hold this off because added code.

mfn:
Hmm. Is there an easier way to achieve this without the this PRs `AliasArguments` and it's dependency? Just a gut feeling but it seems a lot of logic necessary for this.

(note: please don't be offended; I haven't attempted an implementation myself, just thought I'd ask…)

georgeboot:
There might be.

What is your main concern? Added dependency, performance (can’t easily opt-out or this code) or complexity?

mfn:
Haha, all :-)

I think I'm overruled with my performance concerns in that a) there's more need for hit and b) so far the concern is only on paper, no one (including myself) has proved it.

I would definitely prefer a dependency-less solution. Not because I'm against reusing code, but it seems to me this package was created with the almost sole purpose here. I.e. it would make more sense to me to have this in this code base, for easier review and future optimizations.

crissi:
I will include the dependency and refactor a bit too see if we can get closer to an acceptable solution.

mfn:
>  too see if we can get closer to an acceptable solution.

Thank you very much!

I know that some of my feedback here is very … non-argumentative on a technical level; and on top not quite so specific. Would appreciate a take from a (slightly) different angle here; thx.

mfn:
@crissi is this ready for review?

crissi:
> @crissi is this ready for review?

yes

src/Support/Field.php
```diff
@@ -204,6 +205,8 @@ protected function getResolver(): ?Closure
                 };
             }
 
+            $arguments[1] = (new AliasArguments($this->args(), $arguments[1]))->get();
```
mfn: How about this:
- we add a new `protected function aliasArgs(array $args): array`
  ```php
  {
    return (new AliasArguments($this->args(), $arguments[1]))->get();
  }
  ```
- we add a new `protected function getArgs(array $args): array`
  ```php
  {
    return $this->aliasArgs($args);
  }
  ```
- we call it here like this: `$arguments[1] = $this->getArgs($arguments[1]);`

I see the benefit that:
- people can decide what to do with the args on their own
- by providing `aliasArgs` we make it easier to re-use if it people want to add additional steps within their `getArgs()`

tests/Unit/AliasAguments/Stubs/UpdateExampleMutation.php
```diff
@@ -0,0 +1,63 @@
+<?php
```
mfn: declare strict_types missing

tests/Unit/AliasAguments/Stubs/ExampleValidationInputObject.php
```diff
@@ -0,0 +1,38 @@
+<?php
```
mfn: declare strict_types missing

tests/Unit/AliasAguments/Stubs/ExampleType.php
```diff
@@ -0,0 +1,26 @@
+<?php
```
mfn: declare strict_types missing

tests/Unit/AliasAguments/Stubs/ExampleNestedValidationInputObject.php
```diff
@@ -0,0 +1,35 @@
+<?php
```
mfn: declare strict_types missing

tests/Unit/AliasAguments/AliasArgumentsTest.php
```diff
@@ -0,0 +1,130 @@
+<?php
```
mfn: declare strict_types missing

crissi:
@mfn can you look again, I see styleci is still active?!

crissi:
@mfn can you take a look again?

crissi:
styleci still running!?

mfn:
@crissi 
Ok, I guess I now see were the confusion also from the other PR came.

Please remove the changes to please styleci => we can ignore the checks by it until it's removed

sorry for the confusion.

mfn:
Please just remove/revert the commits you made to please StyleCI, that should be the easiest way AFAICS.

crissi:
@mfn should be ready for review now

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ CHANGELOG
 --------------
 ### Added
 - Allow passing through an instance of a `Field` [\#521 / georgeboot](https://github.com/rebing/graphql-laravel/pull/521/files)
+- add the ability to alias query and mutations arguments as well as input objects [\#517 / crissi](https://github.com/rebing/graphql-laravel/pull/517/files)
```
mfn:
```suggestion
- Add the ability to alias query and mutations arguments as well as input objects [\#517 / crissi](https://github.com/rebing/graphql-laravel/pull/517/files)
```

Readme.md
```diff
@@ -101,7 +101,7 @@ To work this around:
 ## Usage
 
 - [Laravel GraphQL](#laravel-graphql)
-    - [Note: these are the docs for 3.*, please see the `v1` branch for the 1.* docs](#note-these-are-the-docs-for-3-please-see-the-v1-branch-for-the-1-docs)
+    - [Note: these are the docs for the current release, please see the `v1` branch for the 1.* docs](#note-these-are-the-docs-for-the-current-release-please-see-the-v1-branch-for-the-1-docs)
```
mfn: How is this diff even possible? The master branch literally says this: https://github.com/rebing/graphql-laravel/commit/30e69d32ada916ae5f7caeca4783c0380f167aa7

crissi: probably my markdown editor that changed that, will revert

mfn:
> I'll take care of the broken tests, this is a travis config issue.

https://github.com/rebing/graphql-laravel/pull/538

Readme.md
mfn: No, I mean the version in master already is at the change so technically this shouldn't be possible.

No need to change it, it's the _correct_ text anyway!

mfn:
Awesome, thank you!

PR #520, opened 2019-11-10
resolve method can use dependency injection

crissi:
- the resolve method now only  have 3 params that are required
- any class can be injected in the resolve method, including the ResolveInfo and SelectFields class
- you no longer need to use $selectfields as a closure, but is still supported for when you want to change the depth

*make resolve works more like the controller methods

### DIFF
--- a/src/Support/Field.php
+++ b/src/Support/Field.php
@@ -14,8 +14,10 @@ use GraphQL\Type\Definition\WrappingType;
 use Illuminate\Contracts\Validation\Validator as ValidatorContract;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Validator;
+use InvalidArgumentException;
 use Rebing\GraphQL\Error\AuthorizationError;
 use Rebing\GraphQL\Error\ValidationError;
+use ReflectionMethod;
 
 /**
  * @property string $name
@@ -203,22 +205,51 @@ abstract class Field
                 }
             }
 
-            // Add the 'selects and relations' feature as 5th arg
-            if (isset($arguments[3])) {
-                $arguments[] = function (int $depth = null) use ($arguments): SelectFields {
-                    $ctx = $arguments[2] ?? null;
-
-                    return new SelectFields($arguments[3], $this->type(), $arguments[1], $depth ?? 5, $ctx);
-                };
-            }
-
             // Authorize
-            if (call_user_func_array($authorize, $arguments) != true) {
+            if (true != call_user_func_array($authorize, $arguments)) {
                 throw new AuthorizationError('Unauthorized');
             }
 
-            return call_user_func_array($resolver, $arguments);
+            $method = new ReflectionMethod($this, 'resolve');
+
+            $additionalParams = array_slice($method->getParameters(), 3);
+
+            $additionalArguments = array_map(function ($param) use ($arguments) {
+                $className = null !== $param->getClass() ? $param->getClass()->getName() : null;
+
+                if (null === $className) {
+                    throw new InvalidArgumentException("'{$param->name}' could not be injected");
+                }
+
+                if (Closure::class === $param->getType()->getName()) {
+                    return function (int $depth = null) use ($arguments): SelectFields {
+                        return $this->instanciateSelectFields($arguments, $depth);
+                    };
+                }
+
+                if (SelectFields::class === $className) {
+                    return $this->instanciateSelectFields($arguments);
+                }
+
+                if (ResolveInfo::class === $className) {
+                    return $arguments[3];
+                }
+
+                return app()->make($className);
+            }, $additionalParams);
+
+            return call_user_func_array($resolver, array_merge(
+                [$arguments[0], $arguments[1], $arguments[2]],
+                $additionalArguments
+            ));
         };
+    }
+
+    private function instanciateSelectFields(array $arguments, int $depth = null): SelectFields
+    {
+        $ctx = $arguments[2] ?? null;
+
+        return new SelectFields($arguments[3], $this->type(), $arguments[1], $depth ?? 5, $ctx);
     }
 
     /**
--- a/tests/Support/Objects/ClassToInject.php
+++ b/tests/Support/Objects/ClassToInject.php
@@ -1,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rebing\GraphQL\Tests\Support\Objects;
+
+class ClassToInject
+{
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 --------------
 ### Added
 - Allow passing through an instance of a `Field` [\#521 / georgeboot](https://github.com/rebing/graphql-laravel/pull/521/files)
+- Classes can now be injected in the Resolve method from the query/mutation similarly to Laravel controller methods [\#520 / crissi](https://github.com/rebing/graphql-laravel/pull/520/files)
 ### Fixed
 - Fix validation rules for non-null list of non-null objects [\#511 / crissi](https://github.com/rebing/graphql-laravel/pull/511/files)
 - Add morph type to returned models [\#503 / crissi](https://github.com/rebing/graphql-laravel/pull/503)
--- a/Readme.md
+++ b/Readme.md
@@ -113,6 +113,7 @@ To work this around:
     - [Creating a mutation](#creating-a-mutation)
       - [Adding validation to a mutation](#adding-validation-to-a-mutation)
       - [File uploads](#file-uploads)
+    - [Resolve method](#resolve-method)
     - [Authorization](#authorization)
     - [Privacy](#privacy)
     - [Query Variables](#query-variables)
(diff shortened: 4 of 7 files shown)


mfn:
Just a few high level feelings, will do a detailed review at a later time:
- I _like the concept_
- I wish, in this case, the container would be already "there" (`app()…`) because I also foresee use cases within other resolvers having access to it directly)
- I'm worried about performance impact (reflection…) and the additional logic I see
- I'm worried about correctness with (apollo) batched requests

mfn:
> I'm worried about performance impact (reflection…) and the additional logic I see

- I picked the test `\Rebing\GraphQL\Tests\Database\SelectFieldsTest::testWithNonNullSelectFieldsAndModel`
- changed the call to this
  ```php
        foreach (range(1, 10000) as $_) {
            $response = $this->call('GET', '/graphql', [
                'query' => $graphql,
            ]);
        }
  ```
- ran it 3 times with master and with this PR

master:
```
Time: 15.07 seconds, Memory: 30.00 MB
Time: 14.39 seconds, Memory: 30.00 MB
Time: 14.49 seconds, Memory: 30.00 MB
```
This PR:
```
Time: 15.37 seconds, Memory: 30.00 MB
Time: 17.97 seconds, Memory: 30.00 MB
Time: 15.63 seconds, Memory: 30.00 MB
```
The observation here is: it's measurable.

mfn:
Another thing / train of thought: whilst convenient, this only affects query/mutation resolvers, not field resolvers.

But I dare not to add more complexity to the resolvers in general. But if within a field resolver one needs a dependency, one has to go back to `app()` again.

I wonder if it would make sense to provide more direct access to the container? E.g. via a property `$laravel` on the field type. Then _any_ resolve can use it with `$this->laravel->make(…)`.

But I'm not sure that's smart either. Containers aren't supposed to be passed around.

crissi:
> Another thing / train of thought: whilst convenient, this only affects query/mutation resolvers, not field resolvers.
> 
the reason why I came up with this a final resolver will act more like a Laravel controller methods where it also possible to inject dependencies

crissi:
> The bindings are global (ResolveInfo, SelectFields), which means even after the resolve is done and are still bound and linger around => I don't think that's a good idea. I think this is a fundamental thing: those args are effectively local data and only make sense _in that specific resolver_ and nowhere else.

Okay I will change that, maybe that will give some more acceptable performance.

mfn:
> the reason why I came up with this a final resolver will act more like a Laravel controller methods where it also possible to inject dependencies

And I took think the approach is right 👍  I don't think having the ability to auto-inject into field resolvers will do us any good in terms of performance.

The biggest result I know/work with returns (worst case) ~1000 models and queries around 20 fields I think. That wouldn't fly :-)

stevelacey:
I am a big fan of this, it'd really help tidy up some queries and mutations my project has going on

stevelacey:
Random: being able to inject the request likely has uses – for example, I'm currently working with phpleague oauth and throttlelogins, both expect the request object a lot.

mfn:
@crissi is this ready for review?

crissi:
> @crissi is this ready for review?

yes

src/Support/Field.php
```diff
@@ -198,24 +200,53 @@ protected function getResolver(): ?Closure
...
+
+            return call_user_func_array($resolver, array_merge(
+                [
+                $arguments[0], $arguments[1], $arguments[2], ],
+                $additionalArguments
+            ));
         };
     }
 
+    private function instanciateSelectFields(array $arguments, $depth = null): SelectFields
```
mfn: type for `$depth` missing

src/Support/Field.php
```diff
@@ -198,24 +200,53 @@ protected function getResolver(): ?Closure
...
+                if ($param->getType()->getName() === 'Closure') {
+                    return function (int $depth = null) use ($arguments) : SelectFields {
+                        return $this->instanciateSelectFields($arguments, $depth);
+                    };
+                }
+
+                $className = $param->getClass()->getName() ?? null;
+                if (is_null($className)) {
+                    throw new InvalidArgumentException("{$param->name} is not a class");
+                }
```
mfn: This check should be at the very top of the closure

src/Support/Field.php
```diff
@@ -198,24 +200,53 @@ protected function getResolver(): ?Closure
...
+
+            $additionalArguments = array_map(function ($param) use ($arguments) {
+                if ($param->getType()->getName() === 'Closure') {
+                    return function (int $depth = null) use ($arguments) : SelectFields {
+                        return $this->instanciateSelectFields($arguments, $depth);
+                    };
+                }
+
+                $className = $param->getClass()->getName() ?? null;
+                if (is_null($className)) {
```
mfn: We only use `null ===` in this lib

src/Support/Field.php
```diff
@@ -198,24 +200,53 @@ protected function getResolver(): ?Closure
...
                 throw new AuthorizationError('Unauthorized');
             }
 
-            return call_user_func_array($resolver, $arguments);
+            $method = new ReflectionMethod($this, 'resolve');
+
+            $additionalParams = array_slice($method->getParameters(), 3);
+
+            $additionalArguments = array_map(function ($param) use ($arguments) {
+                if ($param->getType()->getName() === 'Closure') {
```
mfn: Seeing that the other uses reference FQ class names, this should be `Closure::class` (and be imported) I guess?

src/Support/Field.php
```diff
@@ -198,24 +200,53 @@ protected function getResolver(): ?Closure
...
+                if ($className === ResolveInfo::class) {
+                    return $arguments[3];
+                }
+
+                return app()->make($className);
+            }, $additionalParams);
+
+            return call_user_func_array($resolver, array_merge(
+                [
+                $arguments[0], $arguments[1], $arguments[2], ],
```
mfn: Is it me or is the formatting off here? Can we do:
```php
[$arguments[0], $arguments[1], $arguments[2]],
```

src/Support/Field.php
```diff
@@ -198,24 +200,52 @@ protected function getResolver(): ?Closure
...
 
-            return call_user_func_array($resolver, $arguments);
+            $method = new ReflectionMethod($this, 'resolve');
+
+            $additionalParams = array_slice($method->getParameters(), 3);
+
+            $additionalArguments = array_map(function ($param) use ($arguments) {
+                $className = $param->getClass()->getName() ?? null;
+                if (null === $className) {
+                    throw new InvalidArgumentException("{$param->name} is not a class");
```
mfn: Could you please also add a test for this case?

tests/Support/Objects/ClassToInject.php
```diff
@@ -0,0 +1,7 @@
+<?php
```
mfn: declare strict_types missing

I see the hint we need this on php-cs fixer 😄

crissi:
@mfn can you take a look again?

mfn:
@crissi 
image

I don't even need to take a look to know something's not right :-/

Can you please undo this formatting changes?

mfn:
Ok, sorry, same as the other PR.

Please remove any changes you did to try to please StyleCI. It's still enabled, yes, but we actually only use php-cs fixer already. The reports from StyleCI are not required to be green for merging, so from that workflow side it's ok. It's just confusing as hell until it's completely removed.

Sorry 🤷‍♀️

crissi:
> Ok, sorry, same as the other PR.
> 
> Please remove any changes you did to try to please StyleCI. It's still enabled, yes, but we actually only use php-cs fixer already. The reports from StyleCI are not required to be green for merging, so from that workflow side it's ok. It's just confusing as hell until it's completely removed.
> 
> Sorry 🤷‍♀️

Did run composer fix-style, but apparently had the old phpcs config lying around

crissi:
Running it again correctly, but I still get the same changes, wtf...

mfn:
> Running it again correctly, but I still get the same changes...

I think to fix this should be easy as just removing/reverting all the commits you made specifically for StyleCI

crissi:
@mfn it should be ready for review now

src/Support/Field.php
```diff
@@ -204,26 +206,55 @@ protected function getResolver(): ?Closure
...
-
             $arguments[1] = $this->getArgs($arguments);
 
             // Authorize
-            if (call_user_func_array($authorize, $arguments) != true) {
+            if (true != call_user_func_array($authorize, $arguments)) {
                 throw new AuthorizationError('Unauthorized');
             }
 
-            return call_user_func_array($resolver, $arguments);
```
milewski: Why not just calling the app()->call() directly in here? 

```php
return app()->call([ $this, 'resolve' ], [
   'root' => $root,
    'args' => $args, 
    'ctx' => $ctx,
    'resolveInfo' => $arguments[3],
    'selectFields' => $this->instanciateSelectFields($arguments),
]);
```

mfn: 🤷‍♀️

You can always try making a PR!

mfn: > Would it work?

I can't say, didn't author it /cc @crissi

crissi: seems like it would bound to specific variable names instead of now where only the 3 first are required arguments and you can use what ever variable name there after as long as the Class name is correct.

PR #521, opened 2019-11-11
Allow using instance of Field

georgeboot:
This code enabled you to use an instance of a Field instead of only specifying it's class. This enables some cool reusability scenarios.

The problem we wanted to solve is reusability of both a field definition (with args) but also the resolver.

By default you define your field in a field class and your resolver (mapper) in a scalar.
But since the scalar can't access the `$args`, we opted to do it this way.

We're currently using it for our `FormattableDate` type.

```php
<?php

class Purchase extends GraphQLType
{
    protected $attributes = [
        'name' => 'Purchase',
        'model' => PurchaseModel::class,
    ];

    public function fields(): array
    {
        return [
            // ...

            'createdAt' => new Fields\FormattableDate([
                'alias' => 'created_at',
            ]),
        ];
    }
}
```

```php
<?php

class FormattableDate extends Field
{
    protected $attributes = [
        'name' => 'FormattableDate',
    ];

    public function __construct(array $settings = [])
    {
        $this->attributes = \array_merge($this->attributes, $settings);
    }

    protected function getProperty(): string
    {
        return $this->attributes['alias'] ?? $this->attributes['name'];
    }

    public function args(): array
    {
        return [
            'human' => [
                'type' => Type::boolean(),
                'defaultValue' => false,
            ],
            'format' => [
                'type' => Type::string(),
                'defaultValue' => 'Y-m-d H:i',
                'description' => 'Defaults to Y-m-d H:i',
            ],
            'relative' => [
                'type' => Type::boolean(),
                'defaultValue' => false,
            ],
        ];
    }

    public function type(): Type
    {
        return GraphQL::type('FormattableDate');
    }

    public function resolve(Model $root, array $args): string
    {
        /** @var Carbon $date */
        $date = $root->{$this->getProperty()};

        if ($args['human']) {
            return $date->format('l j F Y');
        }

        if ($args['relative']) {
            return $date->diffForHumans();
        }
        
        return $date->format($args['format']);
    }
}
```


src/Support/Type.php
```diff
@@ -89,6 +89,9 @@ public function getFields(): array
                 $field = app($field);
                 $field->name = $name;
                 $allFields[$name] = $field->toArray();
+            } elseif ($field instanceof Field) {
+                $field->setName($name);
```
mfn: Let's do it without the setter, as it's done 3 lines above => just `$field->name = $name`.

The reason is homogeneity:
- `\Rebing\GraphQL\Support\Field` implements `__set` exactly for this reason.
- there should be one "true" way of doing things
- introducing this would mean changing _all_ of the instances we've (refactor)

If this is so important, it should be done in separate PR

georgeboot: Yeah I initially did that but CI (phpstan) was complaining. So I did it this way.

Will change back, add tests and docs first thing tomorrow. M

georgeboot: Done. PHPstan still complaining though. The class has a `@property` annotation already, so not really sure why.

```
Line   src/Support/Type.php
 ------ ----------------------------------------------------------------------
  93     Access to an undefined property Rebing\GraphQL\Support\Field::$name.
```

PR #530, opened 2019-11-21
Implementing ClientAware on exceptions

georgeboot:
## Summary

Implement the `ClientAware` interface on the `ValidationError` and `AuthorizationError`.

This will add the error category to an error response. Please refer to https://webonyx.github.io/graphql-php/error-handling/.

Technically this could be a breaking change, because error category changes from `graphql` to `validation|authorization`. However, I don't believe that anyone implemented that since the `graphql` category was used everywhere, and it would thus be pointless to check for it.

I also fixed a small typo in the pull request template.

## Type of change

### Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -12,4 +12,4 @@
 
 ### Checklist
 - [ ] Existing tests have been adapted and/or new tests have been added
-- [ ] Code style has been fixed via `composer fix-tyle`
+- [ ] Code style has been fixed via `composer fix-style`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ CHANGELOG
 - Add morph type to returned models [\#503 / crissi](https://github.com/rebing/graphql-laravel/pull/503)
 ### Changed
 - Switch Code Style handling from StyleCI to PHP-CS Fixer [\#502 / crissi](https://github.com/rebing/graphql-laravel/pull/502)
-- Implemented [ClientAware](https://webonyx.github.io/graphql-php/error-handling/#default-error-formatting) interface on integrated exceptions [\#502 / georgeboot](https://github.com/rebing/graphql-laravel/pull/530)
+- Implemented [ClientAware](https://webonyx.github.io/graphql-php/error-handling/#default-error-formatting) interface on integrated exceptions [\#530 / georgeboot](https://github.com/rebing/graphql-laravel/pull/530)
 
 2019-10-23, 3.1.0
 -----------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ CHANGELOG
 - Add morph type to returned models [\#503 / crissi](https://github.com/rebing/graphql-laravel/pull/503)
 ### Changed
 - Switch Code Style handling from StyleCI to PHP-CS Fixer [\#502 / crissi](https://github.com/rebing/graphql-laravel/pull/502)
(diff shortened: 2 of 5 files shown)


mfn:
Oh, and could you please also add a CHANGELOG.md entry under "Changed"?

georgeboot:
> In this case `\Rebing\GraphQL\GraphQL::formatError` should be adapted to make use of the category.

Why does this need to be adjusted? `webonyx/graphql-php` will handle this automatically.

mfn:
> Why does this need to be adjusted? `webonyx/graphql-php` will handle this automatically.

How can it, this is our library here:
image

Or am I missing something?

georgeboot:
Ah yes. Now I see.

The `ClientSafe` interface will only set the `errors.0.category`. This is `GraphQL` by default.

However, in clients, it’s very hard to check what error is actually happening other than comparing error messages. And we all know that is bad.

If we have a category to compare with, it’s much easier. If the category is validation, we need to run logic to handle validation issues etc.

georgeboot:
My custom error handler and formatter:

```php
<?php

declare(strict_types=1);

namespace App\Support;

use GraphQL\Error\Debug;
use GraphQL\Error\Error;
use GraphQL\Error\FormattedError;
use NunoMaduro\Collision\Adapters\Laravel\ExceptionHandler;
use Rebing\GraphQL\Error\ValidationError;
use Throwable;

class GraphQLErrorHandler
{
    public static function formatError(Error $error): array
    {
        $debug = config('app.debug') ? (Debug::INCLUDE_DEBUG_MESSAGE | Debug::INCLUDE_TRACE) : 0;
        $formatter = FormattedError::prepareFormatter(null, $debug);
        
        return $formatter($error);
    }

    public static function handleErrors(array $errors, callable $formatter): array
    {
        $handler = function (Error $error) use ($formatter): array {
            $error = self::prepareError($error);

            return $formatter($error);
        };

        return \array_map($handler, $errors);
    }

    public static function prepareError(Error $error): Error
    {
        $underlyingException = $error->getPrevious();

        if ($underlyingException) {
            $error = self::generateNewErrorBasedOnPreviousException($error, $underlyingException);
        }

        resolve(ExceptionHandler::class)->report($error);

        return $error;
    }

    public static function generateNewErrorBasedOnPreviousException(Error $error, Throwable $previous): Error
    {
        $message = $error->message;
        $nodes = $error->nodes;
        $source = $error->getSource();
        $positions = $error->getPositions();
        $path = $error->getPath();
        $extensionContent = [];

        if ($previous instanceof RendersErrorsExtensions) {
            $extensionContent = $previous->extensionsContent();
        } elseif ($previous instanceof ValidationError) {
            $extensionContent = [
                'validation' => $previous->getValidatorMessages(),
            ];
        }

        return new Error(
            $message,
            $nodes,
            $source,
            $positions,
            $path,
            $previous,
            $extensionContent
        );
    }
}

```

In my exception handler, I've added the following to my '$dontReport`:
```php
protected $dontReport = [
    \Rebing\GraphQL\Error\ValidationError::class,
];
```

mfn:
Oh, I already understand. I kind of treated the graphql category and the extension sub-namespace be the same. In this case, disregard my comment.

Is it possible to add a test, exposing this field?

georgeboot:
> Is it possible to add a test, exposing this field?

Sure, on it.

mfn:
I guess we'll be able to merge this soon, can you please also sync it with master and resolve the conflict?

georgeboot:
Is there any desire to rewrite the error handler/formatter of this package to support extension content etc? I can do a PR for this.

mfn:
> Is there any desire to rewrite the error handler/formatter of this package to support extension content etc? I can do a PR for this.

Not sure what you mean exactly.

Thanks for the PR 😄

PR #545, opened 2019-11-26
wrong return type of getFields in type

crissi:
## Summary

Types does not match usage, see here, it allows numeric int for keys when FieldDefinition:
https://github.com/rebing/graphql-laravel#even-better-reusable-fields

## Type of change

### DIFF
--- a/src/Support/Type.php
+++ b/src/Support/Type.php
@@ -39,7 +39,7 @@ abstract class Type implements TypeConvertible
     }
 
     /**
-     * @return array<string,array|string|FieldDefinition|Field>
+     * @return array<int|string,array|string|FieldDefinition|Field>
      */
     public function fields(): array
     {
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -41,4 +41,6 @@ parameters:
         - '/Method Rebing\\GraphQL\\Tests\\Unit\\UploadTests\\UploadSingleFileMutation::resolve\(\) should return string but returns string\|false/'
         # tests/Database/AuthorizeArgsTests/TestAuthorizationArgsQuery.php
         - '/Trying to invoke Closure\|null but it might not be a callable/'
+        - '/Property Rebing\\GraphQL\\Support\\Field\:\:\$name \(string\) does not accept int\|string\./'
+        - '/Parameter #1 \$name of method Rebing\\GraphQL\\Support\\Type\:\:getFieldResolver\(\) expects string, int\|string given\./'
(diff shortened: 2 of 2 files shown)


src/Support/Type.php
```diff
@@ -39,7 +39,7 @@ public function attributes(): array
     }
 
     /**
-     * @return array<string,array|string|FieldDefinition|Field>
+     * @return array<int|string,array|string|FieldDefinition|Field>
```
mfn: Are you sure? Even in the linked readme example I can't see this. They "key" of fields always has to be a string, i.e. the name / identifier of the field.

crissi: image


these have int as key, since when you do not add a key it will be int

mfn: O_O

mfn: True, in this case, this part in `\Rebing\GraphQL\Support\Type::getFields` "makes it work":
```php
            } elseif ($field instanceof FieldDefinition) {
                $allFields[$field->name] = $field;
            } else {
```

Confusing

mfn: Not sure how to each phpstan properly about this or if we should just ignore the errors.
```
 ------ ---------------------------------------------------------------------------------------------------------------- 
  Line   src/Support/Type.php                                                                                            
 ------ ---------------------------------------------------------------------------------------------------------------- 
  93     Property Rebing\GraphQL\Support\Field::$name (string) does not accept int|string.                               
  98     Parameter #1 $name of method Rebing\GraphQL\Support\Type::getFieldResolver() expects string, int|string given.  
 ------ ---------------------------------------------------------------------------------------------------------------- 
```
The static analyzer can't interfere the context in which we implicitly know that `$name` is an int 🤷‍♀️

mfn: I wonder if it's possible to (re)write the `getFields` method to make this clearer so we don't need to nudge phpstan.

OTOH currently this method looks compact and not-so-bad to read, so we might opt for ignoring the errors. I'm undecided.

Either way, willing to merge this but tests need to be green anyway.

crissi: Phpstan is actually correct in it assumptions because the typehint is still not correct.

since the int key in the return type should only apply when fieldDefinition is there. But don't think that is possible in docblock

I am ignoring the errors

crissi:
waiting for something here?

mfn:
> waiting for something here?

lol, no. At the time I waited for green tests but then got distracted 🤪

PR #546, opened 2019-11-26
fix alias with nullable fields

crissi:
## Summary
Fixes input alias with fields set to null or not set at all

## Type of change

### Checklist

### DIFF
--- a/src/Support/AliasArguments/AliasArguments.php
+++ b/src/Support/AliasArguments/AliasArguments.php
@@ -13,9 +13,6 @@ use GraphQL\Type\Definition\WrappingType;
 
 class AliasArguments
 {
-    private $typedArgs;
-    private $arguments;
-
     public function get(array $typedArgs, array $arguments): array
     {
         $pathsWithAlias = $this->getAliasesInFields($typedArgs, '');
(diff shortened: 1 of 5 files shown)


crissi:
running this though a real life test and discovering some issues

mfn:
Should I still wait for merging or go ahead?

crissi:
should be ready now

src/Support/AliasArguments/ArrayKeyChange.php
```diff
@@ -39,7 +39,7 @@ private function changeKey(array $target, array $segments, string $replaceKey):
         $segment = array_shift($segments);
 
         if (empty($segments)) {
-            if (isset($target[$segment])) {
+            if (array_key_exists($segment, $target) && $replaceKey !== $segment) {
```
mfn: I don't see this change fully covered.

When I test it with an `} else { throw new \RuntimeException('boom'); }`, that exception is never thrown.

Is it possible to cover this?

I checked also the other changes, they are 🤞

crissi: The null_value field I added covers this it would never, alias if not using array_key_exists

mfn: Makes sense, thanks 👍

src/Support/AliasArguments/AliasArguments.php
```diff
@@ -13,9 +13,6 @@
 
 class AliasArguments
 {
-    private $typedArgs;
-    private $arguments;
-
```
mfn: Removed them right in this PR as I noticed they were unused after the code was refactored

PR #552, opened 2019-11-27
[GITHUB] Add workflow for phpstan and remove it from travis

mfn:
## Summary
See how things work out with Github Actions

## Type of change

### TODO
- [x] Caching is missing
  Runtime currently is pretty awful due to this
- [x] Output of composer install step is messy, works much better on travis

### DIFF
--- a/.github/workflows/phpstan.yml
+++ b/.github/workflows/phpstan.yml
@@ -1,0 +1,26 @@
+name: CakePHP CI
+
+on: [push, pull_request]
+
+jobs:
+  static-analysis:
+    name: Static Analysis
+    runs-on: ubuntu-18.04
+
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Setup PHP
+        run: sudo phpdismod xdebug
+
+      - uses: actions/cache@v1
+        with:
+          path: ~/.composer/cache/files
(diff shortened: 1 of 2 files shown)


mfn:
I don't see a "need" for this, I simply found it a nice opportunity to finally play around with Github Actions.

Any comments? Should we do this?

The case can be made that we could move over more and more stuff and basically parallelize between travis and github actions, reducing overall time for all checks.

.github/workflows/phpstan.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Setup PHP
+        run: sudo phpdismod xdebug
+
+      - uses: actions/cache@v1
+        with:
+          path: ~/.composer/cache/files
+          key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
```
mfn: Usually this uses `composer.lock` but, as a library, we don't have such a file.

This also means _any_ change to `composer.json` will invalidate the installation cache. But given that it doesn't change very often, I'd say this is acceptable for a start.

mfn:
Going to merge this now; if we're not happy with this we can always revert it 🤞

PR #578, opened 2020-01-19
Add support for custom authorization message

Sh1d0w:
## Summary
Currently it is not possible to override the authorization message in case the authorization message returns false.

## Type of change

### Checklist
- [x] Update the README.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ CHANGELOG
 [Next release](https://github.com/rebing/graphql-laravel/compare/4.0.0...master)
 --------------
 
+### Added
+- Add support for custom authorization message
+
 2019-12-09, 4.0.0
 -----------------
 ### Added
--- a/src/Support/Field.php
+++ b/src/Support/Field.php
@@ -210,7 +210,7 @@ abstract class Field
 
             // Authorize
             if (true != call_user_func_array($authorize, $arguments)) {
-                throw new AuthorizationError('Unauthorized');
+                throw new AuthorizationError($this->getAuthorizationMessage());
             }
 
             $method = new ReflectionMethod($this, 'resolve');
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ CHANGELOG
 [Next release](https://github.com/rebing/graphql-laravel/compare/4.0.0...master)
 --------------
 
+
+2020-01-19, 4.0.0
+-----------------
```
mfn: Please remove these lines, as the net release date isn't decided yet!

src/Support/Field.php
```diff
@@ -290,6 +290,14 @@ public function getAttributes(): array
         return $attributes;
     }
 
+    /**
+     * @return string
+     */
```
mfn: Please remove the phpdoc, doesn't "add" anything here, thanks!

tests/Unit/GraphQLQueryTest.php
```diff
@@ -120,6 +120,18 @@ public function testQueryAndReturnResultWithAuthorize(): void
         $this->assertEquals('Unauthorized', $result['errors'][0]['message']);
     }
 
+    /**
+     * Test query with authorize.
+     */
+    public function testQueryAndReturnResultWithAuthorizeMessage(): void
```
mfn:
```suggestion
    public function testQueryAndReturnResultWithCustomAuthorizeMessage(): void
```
What do you think about this suggestion?

CHANGELOG.md
Sh1d0w: Sure will do. I haven't added this initially, but saw in the PR checklist it was required to amend CHANGELOG.md and wasn't sure what exactly to add. Are you happy to remove the whole entry I've added?

Sh1d0w:
Thanks for the feedback @mfn , I've updated my PR.

CHANGELOG.md
mfn: No, please keep the changelog (it's a "change log", after all 😜) but only remove the lines I covered in github!

Sh1d0w: @mfn Right, should be ok now I think :)

mfn:

PR #579, opened 2020-01-19
Fix infinite loop

### LINKED ISSUE #512: Infinite loop while input object pointing at each other
https://github.com/Artem-Schander/graphql-infinite-loop

crissi:
## Summary
Fixes #512

Fix the infinite loop as well as sending the correct matching input data to the rule-callback.

If people have been using rule-callback in a nested input object, the input to the object was before always the full input, now the input data matches the level of nesting.

## Type of change

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -446,7 +446,13 @@ class UpdateUserEmailMutation extends Mutation
     {
         return [
             'id' => ['required'],
-            'email' => ['required', 'email']
+            'email' => ['required', 'email'],
+            'password' => function($inputArguments, $mutationArguments) {
+                if ($inputArguments['id'] !== 1337) {
+                    return ['required'];
+                }
+                return [];
+            }
         ];
     }
 
--- a/tests/Unit/MutationTest.php
+++ b/tests/Unit/MutationTest.php
@@ -4,12 +4,13 @@ declare(strict_types=1);
 
 namespace Rebing\GraphQL\Tests\Unit;
 
-use Illuminate\Support\Arr;
 use Illuminate\Validation\Validator;
 use Rebing\GraphQL\Error\ValidationError;
 use Rebing\GraphQL\Tests\Support\Objects\ExampleNestedValidationInputObject;
+use Rebing\GraphQL\Tests\Support\Objects\ExampleRuleTestingInputObject;
 use Rebing\GraphQL\Tests\Support\Objects\ExampleType;
 use Rebing\GraphQL\Tests\Support\Objects\ExampleValidationInputObject;
+use Rebing\GraphQL\Tests\Support\Objects\UpdateExampleMutationForRuleTesting;
 use Rebing\GraphQL\Tests\Support\Objects\UpdateExampleMutationWithInputType;
 
 class MutationTest extends FieldTest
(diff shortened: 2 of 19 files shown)


mfn:
Ok! That one looks massive…
will take me some time to dig through it; thanks!

mfn:
Sorry for the long wait, I've looked into the high level functionality: very cool, totally want to merge this!

Going now through the code for giving detailed feedback.

crissi:
All cool, take your time😊

Sendt: Saturday, January 25, 2020 8:46:00 PM
Emne: Re: [rebing/graphql-laravel] Fix infinite loop (#579)


Sorry for the long wait, I've looked into the high level functionality: very cool, totally want to merge this!

Going now through the code for giving detailed feedback.

tests/Support/Objects/UpdateExampleMutationWithInputType.php
```diff
@@ -49,7 +49,6 @@ public function args(): array
             ],
 
             'test_with_rules_nullable_input_object' => [
-                'name' => 'test',
```
mfn: Why?

tests/Support/Objects/ExampleValidationInputObject.php
```diff
@@ -24,10 +24,15 @@ public function fields(): array
     {
         return [
             'val' => [
-                'name' => 'val',
```
mfn: Why?

tests/Support/Objects/UpdateExampleMutationWithInputType2.php
```diff
@@ -0,0 +1,73 @@
...
+declare(strict_types=1);
+
+namespace Rebing\GraphQL\Tests\Support\Objects;
+
+use GraphQL\Type\Definition\Type;
+use PHPUnit\Framework\Assert;
+use Rebing\GraphQL\Support\Facades\GraphQL;
+use Rebing\GraphQL\Support\Mutation;
+
+class UpdateExampleMutationWithInputType2 extends Mutation
```
mfn: `2`, same feedback as the other one

Readme.md
```diff
@@ -446,7 +446,13 @@ class UpdateUserEmailMutation extends Mutation
     {
         return [
             'id' => ['required'],
-            'email' => ['required', 'email']
+            'email' => ['required', 'email'],
+            'password' => function($arguments, $allArguments) {
```
mfn: `$arguments, $allArguments`

I guess this comes from `\Rebing\GraphQL\Support\Rules::resolveRules`

Do we have better names for this (also through the code / tests), this is quite ambiguous to say at least.

tests/Unit/MutationTest.php
```diff
@@ -123,6 +92,36 @@ public function testValidationError(): void
 
         $attributes = $field->getAttributes();
 
+        try {
+            $attributes['resolve'](null, [
+                'test_with_rules_non_nullable_input_object' => [
+                    'val' => 4,
+                ],
+            ], [], null);
+        } catch (ValidationError $e) {
```
mfn: This pattern is prone to be miss if the underlying code changes/breaks.

I recommend this:
```php
/** @var ValidationError|null $exc */
$exc = null;
try {
  …
} catch (ValidationError $exc) {
  // Deliberately do nothing
}
$this->assertInstanceOf(ValidationError::class, $exc);
// your regular assertions
```

tests/Unit/MutationTest.php
```diff
@@ -123,6 +92,36 @@ public function testValidationError(): void
...
+    public function testWithEmptyInput(): void
+    {
+        $class = $this->getFieldClass();
+        $field = new $class();
+
+        $attributes = $field->getAttributes();
+
         try {
             $attributes['resolve'](null, [], [], null);
         } catch (ValidationError $e) {
```
mfn: I see. The feedback from the last test: this pattern was already here.

Please fix it here, too.

tests/Unit/MutationTest.php
```diff
@@ -131,15 +130,133 @@ public function testValidationError(): void
...
+        $class = $this->getFieldClass();
+        $field = new $class();
+
+        $attributes = $field->getAttributes();
+
+        try {
+            $attributes['resolve'](null, [
+                'test_with_rules' => 'test',
+            ], [], null);
+        } catch (ValidationError $e) {
```
mfn: Please change the pattern here

tests/Unit/MutationTest.php
```diff
@@ -131,15 +130,133 @@ public function testValidationError(): void
...
+        $field = new $class();
+
+        $attributes = $field->getAttributes();
+
+        try {
+            $attributes['resolve'](null, [
+                'test_with_rules_non_nullable_input_object' => [],
+                'test_with_rules_nullable_input_object' => [],
+            ], [], null);
+        } catch (ValidationError $e) {
```
mfn: Please change the pattern here

tests/Unit/MutationTest.php
```diff
@@ -131,15 +130,133 @@ public function testValidationError(): void
...
+        $class = $this->getFieldClass();
+        $field = new $class();
+
+        $attributes = $field->getAttributes();
+
+        try {
+            $attributes['resolve'](null, [
+                'test_with_rules_non_nullable_list_of_non_nullable_input_object' => [],
+            ], [], null);
+        } catch (ValidationError $e) {
```
mfn: Please change the pattern here

tests/Unit/MutationTest.php
```diff
@@ -131,15 +130,133 @@ public function testValidationError(): void
...
+
+        try {
+            $attributes['resolve'](null, [
+                'test_with_rules_non_nullable_list_of_non_nullable_input_object' => [
+                    [
+                        'val' => 1245,
+                    ],
+                ],
+            ], [], null);
+        } catch (ValidationError $e) {
```
mfn: Please change the pattern here

tests/Unit/MutationTest.php
```diff
@@ -169,4 +286,19 @@ public function testCustomValidationErrorMessages(): void
...
+        $attributes = $field->getAttributes();
+
+        try {
+            $attributes['resolve'](null, [
+                'test_with_rules_callback_params' => [
+                    'otherValue' => 1337,
+                ],
+            ], [], null);
+        } catch (ValidationError $e) {
+        }
```
mfn: If we don't care about the exceptions details (beyond the message), I recommend using the built-in phpunit-way: `$this->expectException(…);`

tests/Support/Objects/UpdateExampleMutationWithInputType.php
crissi: don't know why they are there in the first place? does the name attribute do anything?

mfn: > does the name attribute to anything

Well, yes 😄

You can use it to change the name from the key:
- key is a string, use an actual different name
- no explicit key (i.e. plain array), in which case the `name` is actually required

Back to this code: sure, might not be in use/make sense here but my point is: a PR should be focused and random clean ups should not be part of bigger feature like this.

We can totally remove all the `name` which aren't in use, probably a copypasta galore, but then it should be done a) consequently and b) separate from this PR.

crissi:
updated

mfn:

PR #588, opened 2020-02-08
Remove return type from SelectFields::getPrimaryKeyFromParentType()

dharkness:
While `SelectFields::getSelectableFieldsAndRelations()` handles a composite primary key array, the `?string` return type on `getPrimaryKeyFromParentType()` causes a fatal error.

## Summary

Using `HasCompositePrimaryKey` from LaravelTreats allows composite primary keys, but the return type on `getPrimaryKeyFromParentType()` is incompatible with an array. Removing the return type declaration fixes the fatal error and allows querying such a model.

## Type of change

### Checklist
- [ ] Update the README.md

### DIFF
--- a/src/Support/SelectFields.php
+++ b/src/Support/SelectFields.php
@@ -121,7 +121,7 @@ class SelectFields
         return isset($parentType->config['model']) ? app($parentType->config['model'])->getTable() : null;
     }
 
-    private static function getPrimaryKeyFromParentType(GraphqlType $parentType): ?string
+    private static function getPrimaryKeyFromParentType(GraphqlType $parentType)
     {
         return isset($parentType->config['model']) ? app($parentType->config['model'])->getKeyName() : null;
     }


mfn:
I can't accept any "fixes" for something not supported but Laravel but enabled by a 3rd party package and having no tests.

We had this feature in the past, it broke, was fixed gain, lo' and behold it broke again.

dharkness:
I can understand your hesitation to change the return type annotation. But since the caller of that function supports receiving an array of string keys—and works fine when it does—I figured the `?string` return type was simply a mistake. The feature works fine; it's just the type annotation that is in error.

I will make time to write a test this week if that will help move this forward.

mfn:
> I figured the `?string` return type was simply a mistake

I don't think in this case it was…

> But since the caller of that function supports receiving an array of string keys

I couldn't follow that one, can you be more specific what you mean (links to code or something).

Also, I want to note that the definition of `\Illuminate\Database\Eloquent\Model::getKeyName` currently is (L6):
```php
    /**
     * Get the primary key for the model.
     *
     * @return string
     */
    public function getKeyName()
    {
        return $this->primaryKey;
    }
```
Same goes for `\Illuminate\Database\Eloquent\Model::$primaryKey`.

I'm perfectly there are packages working this around, just wanted to give some insights for the decision process that went into tighten/probably breaking things here.

> I will make time to write a test this week if that will help move this forward.

That will certainly help making a better decision I guess!

dharkness:
The one call to `SelectFields::getPrimaryKeyFromParentType()` is in the same class, `getSelectableFieldsAndRelations()`. If that method receives an array, it treats it as a composite primary key and adds each field to the select list.
```
public static function getSelectableFieldsAndRelations(array $requestedFields, GraphqlType $parentType, ?Closure $customQuery = null, bool $topLevel = true)
{
    ...
    $primaryKey = self::getPrimaryKeyFromParentType($parentType);
    ...
    // If a primary key is given, but not in the selects, add it
    if (! is_null($primaryKey)) {
        if (is_array($primaryKey)) {                           <-- handle composite PK
            foreach ($primaryKey as $key) {
                $select[] = $parentTable ? ($parentTable.'.'.$key) : $key;
            }
        } else {
            ... handle single-column PK ...
        }
    }
```

mfn:
Strange, we must be looking at different code versions, latest version in master looks like this:
```php
    public static function getSelectableFieldsAndRelations(
        array $queryArgs,
        array $requestedFields,
        GraphqlType $parentType,
        ?Closure $customQuery = null,
        bool $topLevel = true,
        $ctx = null
    ) {
        $select = [];
        $with = [];

        if ($parentType instanceof WrappingType) {
            $parentType = $parentType->getWrappedType(true);
        }
        $parentTable = self::getTableNameFromParentType($parentType);
        $primaryKey = self::getPrimaryKeyFromParentType($parentType);

        self::handleFields($queryArgs, $requestedFields, $parentType, $select, $with, $ctx);

        // If a primary key is given, but not in the selects, add it
        if (null !== $primaryKey) {
            $primaryKey = $parentTable ? ($parentTable.'.'.$primaryKey) : $primaryKey;
            if (! in_array($primaryKey, $select)) {
                $select[] = $primaryKey;
            }
        }
```
Which commit hash are you looking at?

dharkness:
Heh, that explains a lot. :) I was using a version from June: `e3d2440b81167b3ce0020fd096df74e7cdde8647`.

However, I already got the sense that this was going to be a no-go so I spent today using the deferred feature to work around the issue. It's not pretty, but it works. Them's the breaks when you have to work with a third-party database structure. :)

Thanks for your time and consideration.

PR #592, opened 2020-02-26
GraphQL Macroable

stevelacey:
## Summary

Adds macro support to the GraphQL service/facade. I am often finding myself wanting Arr/Str-esque helpers for GraphQL behaviours (like spinning up Type's), and given static calls to the GraphQL facade are already used in lots of places, this seems to fit well.

## Type of change

### Checklist
- [x] Update the README.md

Let me know if you need anything I've crossed out.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ CHANGELOG
 
 ### Added
 - Add support for custom authorization message [\#578 / Sh1d0w](https://github.com/rebing/graphql-laravel/pull/578)
+- Add support for macros on the GraphQL service/facade [\#592 / stevelacey](https://github.com/rebing/graphql-laravel/pull/592)
 ### Fixed
 - Fix the infinite loop as well as sending the correct matching input data to the rule-callback [\#579 / crissi](https://github.com/rebing/graphql-laravel/pull/579)
 ### Changed
--- a/README.md
+++ b/README.md
@@ -116,23 +116,23 @@ To work this around:
     - [Resolve method](#resolve-method)
     - [Authorization](#authorization)
     - [Privacy](#privacy)
-    - [Query Variables](#query-variables)
+    - [Query variables](#query-variables)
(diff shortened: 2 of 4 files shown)


stevelacey:
A simple example of where this would be useful, I've inherited a GraphQL API where the type returned can change based on conditions (don't ask); the codebase has lots of this:

```php
public function type() : Type
{
    return $this->hasPagination()
        ? GraphQL::paginate('User')
        : Type::listOf(GraphQL::type('User'));
}
```

It'd be great to wrap this up into a macro like this:

```php
GraphQL::macro('paginateOrListOf', function (string $name) : Type {
    return Something::hasPagination()
        ? GraphQL::paginate($name)
        : Type::listOf(GraphQL::type($name));
});
```

So my queries can do:

```php
public function type() : Type
{
    return GraphQL::paginateOrListOf('User');
}
```

Without having to share that logic via inheritence, traits, or some other class.

This allows the end user to expand upon the provided statics like `GraphQL::type($name)`.

mfn:
It's my personal opinion that Macroable that way is an abomination for proper software development, i.e. what you said would be the "correct way"
> Without having to share that logic via inheritence, traits, or some other class.

But I see the signs (read: pressure from the community) so I'm not vetoing it (just dying a little inside). So let's move forward 🕳

What you be so kind and:
- mention it in the readme with some example code
- add a test for it?
  I don't want to see myself trying to remove this in a commit silently without anyone noticing at 3am in the morning. 😏 

Btw, very nice first contribution ❤️

stevelacey:
@mfn awesome 👏 I'll do those things, and your objection will be noted in the ship's log 🙃

stevelacey:
@mfn readme updated and test added. For the readme I copied and reworded https://laravel.com/docs/6.x/responses#response-macros

Also submitted some unrelated readme fixes that this is based against in #593

mfn:

PR #593, opened 2020-02-26
Readme tweaks

stevelacey:
## Summary

- Moves the `Field deprecation` and `Default field resolver` guides out of `JSON columns`; can only assume this was a mistake when the advanced docs were merged into the readme, these aren't anything to do with JSON columns
- Renamed `Input Alias` to `Field and input alias` given `alias` is supported on both and probably used way more on fields than inputs (news to me it even works on inputs)
- Standardizes README filename casing to match CHANGELOG / other projects
- Standardizes heading casing, seems sentence case was used more than title case so I went with that
- Strips trailing whitespace because editor

### Checklist
- [x] Update the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,23 +116,23 @@ To work this around:
     - [Resolve method](#resolve-method)
     - [Authorization](#authorization)
     - [Privacy](#privacy)
-    - [Query Variables](#query-variables)
+    - [Query variables](#query-variables)
     - [Custom field](#custom-field)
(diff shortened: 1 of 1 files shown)


mfn:
Hah. Did GH realize these commits where contained in the other PR and auto-closed this one as merged?

stevelacey:
@mfn yep, exactly that, I rebased #592 so that it would (and because I didn't want conflicts) 🙃

PR #594, opened 2020-02-27
Resolver middleware

### LINKED ISSUE #591: Resolver Middleware
GraphQL-aware middleware solve a variety of problems for me, a while ago I needed to be able to Logstash various stats about the queries going through my API (e.g. what args were passed, what fields were requested), and more recently I wanted to move authorization ahead of validation after I learnt about https://github.com/rebing/graphql-laravel/issues/481.

Turns out pipelines can do a lot of the heavy lifting here, this is what I'm running with currently, it'd be interesting to wrap this up into native support.

**Middleware resolver:**

```php
<?php

namespace App\GraphQL\Traits;

use Closure;
use App\GraphQL\Middleware;
use Illuminate\Pipeline\Pipeline;
use Illuminate\Support\Facades\App;

trait HasMiddleware
{
    protected $middleware = [
        Middleware\Authenticate::class,
        Middleware\Logstash::class,
        Middleware\Paginator::class,
    ];

    protected function getResolver() : ?Closure
    {
        $resolver = parent::getResolver();

        if (!$resolver) {
            return null;
        }

        return function ($root, ...$arguments) use ($resolver) {
            return app(Pipeline::class)
                ->send(array_merge([$this], $arguments))
                ->through($this->middleware)
                ->via('resolve')
                ->then(function ($arguments) use ($resolver, $root) {
                    $result = $resolver($root, ...array_slice($arguments, 1));

                    App::terminating(function () use ($arguments, $result) {
                        foreach ($this->middleware as $name) {
                            $instance = App::make($name);

                            if (method_exists($instance, 'terminate')) {
                                $instance->terminate($this, ...array_slice($arguments, 1), ...[$result]);
                            }
                        }
                    });

                    return $result;
                });
        };
    }
}
```

**Middleware base class:**

```php
<?php

namespace App\GraphQL\Middleware;

use Closure;

abstract class Middleware
{
    public function resolve(array $arguments, Closure $next)
    {
        return $this->handle(...$arguments, ...[
            function (...$arguments) use ($next) {
                return $next($arguments);
            }
        ]);
    }
}
```

**Example middleware:**

```php
<?php

namespace App\GraphQL\Middleware;

use Closure;
use GraphQL\Type\Definition\ResolveInfo;
use Illuminate;

class Paginator extends Middleware
{
    public function handle($root, $args, $context, ResolveInfo $info, Closure $next)
    {
        Illuminate\Pagination\Paginator::currentPageResolver(function () use ($args) {
            return $args['pagination']['page'] ?? 1;
        });

        return $next($root, $args, $context, $info);
    }

    // public function terminate($root, $args, $context, ResolveInfo $info, $result) {
    //     // optional
    // }
}
```

stevelacey:
## Summary

GraphQL-aware middleware solve a variety of problems, a while ago I needed to be able to Logstash various stats about the queries going through my API (e.g. what args were passed, what fields were requested), and more recently I wanted to move authorization ahead of validation after I learnt about #481. Resolves #591.

## Type of change

### Checklist
- [x] Update the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -771,7 +771,7 @@ To create a new middleware, use the `make:graphql:middleware` Artisan command
 php artisan make:graphql:middleware ResolvePage
 ```
 
-This command will place a new ResolvePage class within your app/Http/Middleware directory.
+This command will place a new ResolvePage class within your app/GraphQL/Middleware directory.
 In this middleware, we will set the Paginator current page to the argument we accept via our `PaginationType`:
 
 ```php
--- a/README.md
+++ b/README.md
@@ -644,14 +644,14 @@ Note: You can test your file upload implementation using [Altair](https://altair
         bodyFormData.set('operationName', null);
         bodyFormData.set('map', JSON.stringify({"file":["variables.file"]}));
         bodyFormData.append('file', this.file);
-        
+
         // Post the request to GraphQL controller
         let res = await axios.post('/graphql', bodyFormData, {
           headers: {
             "Content-Type": "multipart/form-data"
           }
         });
-        
+
         if (res.data.status.code == 200) {
           // On success file upload
           this.file = null;
@@ -832,6 +832,15 @@ abstract class Query extends BaseQuery
         Middleware\ResolvePage::class,
     ];
 }
+```
+
+Alternatively, you can override `getMiddleware` to supply your own logic:
+
+```php
+    protected function getMiddleware(): array
+    {
+        return array_merge([...], $this->middleware);
+    }
 ```
 
 #### Terminable middleware
--- a/README.md
+++ b/README.md
@@ -109,6 +109,10 @@ To work this around:
       - [Adding validation to a mutation](#adding-validation-to-a-mutation)
       - [File uploads](#file-uploads)
     - [Resolve method](#resolve-method)
+    - [Resolver middleware](#resolver-middleware)
+      - [Defining middleware](#defining-middleware)
+      - [Registering middleware](#registering-middleware)
+      - [Terminable middleware](#terminable-middleware)
     - [Authorization](#authorization)
     - [Privacy](#privacy)
     - [Query variables](#query-variables)
@@ -757,6 +761,150 @@ class UsersQuery extends Query
 }
 ```
 
+### Resolver middleware
+
+#### Defining middleware
+
+To create a new middleware, use the `make:graphql:middleware` Artisan command
+
+```sh
+php artisan make:graphql:middleware ResolvePage
+```
+
+This command will place a new ResolvePage class within your app/Http/Middleware directory.
+In this middleware, we will set the Paginator current page to the argument we accept via our `PaginationType`:
+
+```php
+namespace App\GraphQL\Middleware;
+
+use Closure;
+use GraphQL\Type\Definition\ResolveInfo;
+use Illuminate\Pagination\Paginator;
+use Rebing\GraphQL\Support\Middleware;
+
+class ResolvePage extends Middleware
+{
+    public function handle($root, $args, $context, ResolveInfo $info, Closure $next)
+    {
+        Paginator::currentPageResolver(function () use ($args) {
+            return $args['pagination']['page'] ?? 1;
+        });
+
+        return $next($root, $args, $context, $info);
+    }
+}
+```
+
+#### Registering middleware
+
+If you would like to assign middleware to specific queries/mutations,
+list the middleware class in the `$middleware` property of your query class.
+
+```php
+namespace App\GraphQL\Queries;
+
+use App\GraphQL\Middleware;
+use Rebing\GraphQL\Support\Query;
+use Rebing\GraphQL\Support\Query;
+
+class UsersQuery extends Query
+{
+    protected $middleware = [
+        Middleware\Logstash::class,
+        Middleware\ResolvePage::class,
+    ];
+}
+```
+
+If you want a middleware to run during every GraphQL query/mutation to your application,
+list the middleware class in the `$middleware` property of your base query class.
+
+```php
+namespace App\GraphQL\Queries;
+
+use App\GraphQL\Middleware;
+use Rebing\GraphQL\Support\Query as BaseQuery;
+
+abstract class Query extends BaseQuery
+{
+    protected $middleware = [
+        Middleware\Logstash::class,
+        Middleware\ResolvePage::class,
+    ];
+}
+```
+
+#### Terminable middleware
+
+Sometimes a middleware may need to do some work after the response has been sent to the browser.
+If you define a terminate method on your middleware and your web server is using FastCGI,
+the terminate method will automatically be called after the response is sent to the browser:
+
+```php
+namespace App\GraphQL\Middleware;
+
+use Countable;
+use GraphQL\Language\Printer;
+use GraphQL\Type\Definition\ResolveInfo;
+use Illuminate\Contracts\Pagination\LengthAwarePaginator;
+use Illuminate\Pagination\AbstractPaginator;
+use Illuminate\Support\Arr;
+use Illuminate\Support\Facades\Log;
+use Illuminate\Support\Facades\Route;
+use Rebing\GraphQL\Support\Middleware;
+
+class Logstash extends Middleware
+{
+    public function terminate($root, $args, $context, ResolveInfo $info, $result): void
+    {
+        Log::channel('logstash')->info('', (
+            collect([
+                'query' => $info->fieldName,
+                'operation' => $info->operation->name->value ?? null,
+                'type' => $info->operation->operation,
+                'fields' => array_keys(Arr::dot($info->getFieldSelection($depth = PHP_INT_MAX))),
+                'schema' => Arr::first(Route::current()->parameters()) ?? config('graphql.default_schema'),
+                'vars' => $this->formatVariableDefinitions($info->operation->variableDefinitions),
+            ])
+                ->when($result instanceof Countable, function ($metadata) use ($result) {
+                    return $metadata->put('count', $result->count());
+                })
+                ->when($result instanceof AbstractPaginator, function ($metadata) use ($result) {
+                    return $metadata->put('per_page', $result->perPage());
+                })
+                ->when($result instanceof LengthAwarePaginator, function ($metadata) use ($result) {
+                    return $metadata->put('total', $result->total());
+                })
+                ->merge($this->formatArguments($args))
+                ->toArray()
+        ));
+    }
+
+    private function formatArguments(array $args): array
+    {
+        return collect(Arr::sanitize($args))
+            ->mapWithKeys(function ($value, $key) {
+                return ["\${$key}" => $value];
+            })
+            ->toArray();
+    }
+
+    private function formatVariableDefinitions(?iterable $variableDefinitions = []): array
+    {
+        return collect($variableDefinitions)
+            ->map(function ($def) {
+                return Printer::doPrint($def);
+            })
+            ->toArray();
+    }
+}
+```
+
+The terminate method receives both the resolver arguments and the query result.
+
+Once you have defined a terminable middleware, you should add it to the list of
+middleware in your queries and mutations.
+
 ### Authorization
 
 For authorization similar to Laravel's Request (or middleware) functionality, we can override the `authorize()` function in a Query or Mutation.
@@ -1330,7 +1478,9 @@ class UserType extends GraphQLType
 ### Pagination
 
 Pagination will be used, if a query or mutation returns a `PaginationType`.
-Note that you have to manually handle the limit and page values:
+
+Note that unless you use [resolver middleware](#defining-middleware),
+you will have to manually supply both the limit and page values:
 
 ```php
 namespace App\GraphQL\Queries;
--- a/src/Support/Field.php
+++ b/src/Support/Field.php
@@ -121,7 +121,7 @@ abstract class Field
         return Validator::make($args, $rules, $messages);
     }
 
-    protected function getResolver() : ?Closure
+    protected function getResolver(): ?Closure
     {
         $resolver = $this->originalResolver();
 
--- a/src/Support/Field.php
+++ b/src/Support/Field.php
@@ -6,8 +6,10 @@ namespace Rebing\GraphQL\Support;
 
 use Closure;
 use GraphQL\Type\Definition\ResolveInfo;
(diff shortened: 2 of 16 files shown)


mfn:
Looks promising, can't wait for the tests.

I'm also interested how well this works together with "batched requests". Would be great to see this in tests.

Example scenario: a batch request with 2 graphql queries were the middleware decides on the first to skip it but the 2nd should still be resolved correctly.

mfn:
Small feedback: I'm all for consistency and formatting, but would be great if you could resist the urge to mix this in a PR introduces a new feature 😬

stevelacey:
@mfn let me know what you think, I'll add tests once we're happy with the implementation.

I figure I should probably add another way to define global middleware, I for one am using 2 base classes (one for query and one for mutation), and am sharing the `middleware` attribute definition via a trait, which is awkward, and doesn't seem to work once `Field` has a definition either.

Maybe `config`? Or `GraphQL::middleware([])`?

There's also some work to do here to distinguish the config of HTTP middleware.

stevelacey:
> Small feedback: I'm all for consistency and formatting, but would be great if you could resist the urge to mix this in a PR introducing a new feature 😬

Yah trying, every time I come to copy an existing example I find 20 variations 🙊

stevelacey:
@mfn separated out the README fixes https://github.com/rebing/graphql-laravel/pull/595

stevelacey:
@mfn added some tests for an `ExampleMiddleware` that does some things + the command

mfn:
@stevelacey much appreciated!

I'm a bit busy at the moment I would like to take my time with the review, also play around a bit because I might see also some use-cases I want to try out.

Definitely helps to already have tests in this stage to accelerate this 👍 

Please bear with me.

stevelacey:
@mfn no rush, please do take your time 👍

README.md
```diff
@@ -662,6 +666,150 @@ class UsersQuery extends Query
...
+
+#### Defining middleware
+
+To create a new middleware, use the `make:graphql:middleware` Artisan command
+
+```sh
+php artisan make:graphql:middleware ResolvePage
+```
+
+This command will place a new ResolvePage class within your app/Http/Middleware directory.
```
mfn: I think it's in a different dir/namespace
```suggestion
This command will place a new ResolvePage class within your app/GraphQL/Middleware directory.
```

src/Support/Field.php
```diff
@@ -22,6 +24,9 @@ abstract class Field
 {
     protected $attributes = [];
 
+    /** @var string[] */
```
mfn: Please also add `@phpstan-var class-string[]`, thanks!

src/Support/Field.php
```diff
@@ -94,7 +99,38 @@ public function getValidator(array $args, array $rules): ValidatorContract
...
+        return function ($root, ...$arguments) use ($resolver) {
+            return app(Pipeline::class)
+                ->send(array_merge([$this], $arguments))
+                ->through($this->middleware)
+                ->via('resolve')
+                ->then(function ($arguments) use ($resolver, $root) {
+                    $result = $resolver($root, ...array_slice($arguments, 1));
+
+                    foreach ($this->middleware as $name) {
+                        $middleware = App::make($name);
```
mfn: Why are we using `app()` above and `App::make` here? (feels inconsistent)

I guess using the Facade would be preferred? (aka death to global functions)

src/Support/Middleware.php
```diff
@@ -0,0 +1,25 @@
...
+use GraphQL\Type\Definition\ResolveInfo;
+
+abstract class Middleware
+{
+    public function handle($root, $args, $context, ResolveInfo $info, Closure $next)
+    {
+        return $next($root, $args, $context, $info);
+    }
+
+    public function resolve(array $arguments, Closure $next)
```
mfn: Please add a phpdoc that this is called from `\Rebing\GraphQL\Support\Field::getResolver` (no one can statically analyze due to how the code has to work with the string name of this method :/)

src/Support/Field.php
```diff
@@ -94,7 +99,38 @@ public function getValidator(array $args, array $rules): ValidatorContract
         return Validator::make($args, $rules, $messages);
     }
 
-    protected function getResolver(): ?Closure
+    protected function getResolver() : ?Closure
```
mfn:
```suggestion
    protected function getResolver(): ?Closure
```

Guess that's a slap for php-cs-fixer, need to look into this

tests/Support/Objects/ExampleMiddleware.php
```diff
@@ -0,0 +1,42 @@
...
+class ExampleMiddleware extends Middleware
+{
+    public function handle($root, $args, $context, ResolveInfo $info, Closure $next)
+    {
+        if ($args['index'] === 4) {
+            $args['index'] = 0;
+        }
+
+        if ($args['index'] === 5) {
+            throw new \Exception('Index 5 is not allowed');
```
mfn: Please always import, same also below

src/Support/Field.php
stevelacey: Yah I just copied this out of Laravel and never thought about it, facade it is.

src/Support/Field.php
stevelacey: I don't understand what this means, are you suggesting something like this?:

```suggestion
    /** @var class-string<Middleware> */
```

src/Support/Middleware.php
stevelacey: Something like this?

```suggestion
    /**
     * @see Field::getResolver()  Middleware is resolved in the field resolver pipeline
     */
    public function resolve(array $arguments, Closure $next)
```

stevelacey: A related thing I considered was swapping the `handle` and `resolve` names around i.e. so the base class has `handle` that calls `resolve` instead of the current implementation `resolve` that calls `handle` – the default `via` pipelines use is `handle`.

I did it this way round so that end-user GraphQL middleware are more consistent with HTTP (by having a `handle` method) but I'm not sure if that's as useful as them having a `resolve` method to make it clear they intercept resolution rather than the HTTP stack.

stevelacey:
@mfn addressed your comments and added some questions.

Yes, am interested in config/global definition, and maybe even better support for per-query middleware ala HTTP controller constructor `$this->middleware()` calls.

I didn't do that yet because I wondered if config was the right place for it, and because naming things is hard. HTTP middleware lives on the HTTP kernel, the best approximation of that might be adding it onto a `GraphQLServiceProvider`, which also mitigates the conflict with the existing HTTP middleware support in config somewhat. What do you think?

I figure the resolution order would be config, query class attr, constructor, and that'd be good enough for pretty much everyone. Laravel middleware does have the middleware ordering support as an additional feature they added later, we could add that if anyone ever wants it but I wouldn't say it's necessary for release, I've only ever really seen it used for auth anyway, which most will be doing at HTTP level.

src/Support/Field.php
mfn: Not replacing, but in addition ("also add"):
```php
@phpstan-var class-string[]
```
Actually I wouldn't add `Middleware` as a class, but technically we're doing duck-typing here => no one enforces that this has to be extending the abstract class we provide.

src/Support/Middleware.php
mfn: TBH that thought didn't cross my mind.

In fact it felt "natural" that it's the `handle` method one has to implement 🤷‍♀️

mfn: Yeah, the `@see` seems appropriate!

mfn:
> but I wouldn't say it's necessary for release,

Ack 👍 and thanks for your insights!

mfn:
@stevelacey I think I answered all question, from my side it's ready to finalize this!

src/Support/Field.php
stevelacey: I see. So like this? I see no examples in the project of any annotation like this

```suggestion
    /** 
     * @var string[]
     * @phpstan-var class-string[]
     */
```

src/Support/Middleware.php
stevelacey: Handle makes sense to me too, I am happy to run with that

stevelacey:
@mfn what do you think about constructor level middleware additions (`$this->middleware(...)`) and global definition via service provider or similar instead of config? I feel like middleware refs belong more in app code than they do config, config is the one place Laravel doesn't put them.

src/Support/Field.php
mfn: Exactly!

mfn:
Wonder what's going on here 🤔
image

mfn:
> what do you think about constructor level middleware additions (`$this->middleware(...)`) and global definition via service provider or similar instead of config? I feel like middleware refs belong more in app code than they do config, config is the one place Laravel doesn't put them.

I think I would see some more specific proposal or sample code to "get a feel" for this.

middleware "config" is certainly not a Laravel pattern, but OTOH the GraphQL integration absolutely depends on the config.

I mean this could bet yet-another-things to consider: the GraphQL library suggests default paths/namespaces (`app/GraphQL/…`) so we would someone have to list all the types and mutations manually? Why wouldn't there be a mechanism which, like the console commands, can recursively load them all (if desired).

Certainly in my projects that would be welcome, my biggest has over 200 of such registered classes, all manually placed.

However: unless you want to make it the scope of this PR, I suggest to follow up on this rather and finalize this one.

mfn:
One use-case for this would be provide some kind of "filtering" based on permissions.

Ideally this would already be reflected in the schema but that's not always easily possible.

E.g. depending on user permissions, filtering out access to certain queries and mutations via this.

The `handle` method does not receive the query/mutation name itself but AFAIK it's available via `ResolveInfo` which should be good enough:
```php
class ExampleMiddleware extends Middleware
{
    public function handle($root, $args, $context, ResolveInfo $info, Closure $next)
    {
        var_dump($info->fieldName);
```
and the test output:
```
…
string(18) "examplesMiddleware"


Time: 290 ms, Memory: 14.00 MB

OK (1 test, 2 assertions)
```

mfn:
@stevelacey are you planning on moving forward?

Hope everything's fine due to current situation 🤞 

I'm also interested in the PR and I think everything is in place and I could even finish up the PR on my own but didn't want to just take it over.

stevelacey:
> I mean this could bet yet-another-things to consider: the GraphQL library suggests default paths/namespaces (`app/GraphQL/…`) so we would someone have to list all the types and mutations manually? Why wouldn't there be a mechanism which, like the console commands, can recursively load them all (if desired).

Yeah I'd like this too, again, a job for a service provider I think, we could copy the patterns established in `EventServiceProvider` aka `shouldDiscoverEvents` / `discoverEventsWithin`.

Another reason type discovery would be of benefit is by removing the problems created by the config pattern accepting array keys that must then be used when calling `GraphQL::type()`, instead of the type's actual name. This can create a real mess. I think a new type discovery shouldn't support this.

Something to think about later.

> One use-case for this would be provide some kind of "filtering" based on permissions.

Yeah kind of, though they'd still reflect in the introspection, it's not like they'd be truely filtered – authentication/authorization is one of my primary use-cases of middleware though.

> The `handle` method does not receive the query/mutation name itself but AFAIK it's available via `ResolveInfo` which should be good enough:

The `handle` method also receives the query/mutation instance as `$root`.

> @stevelacey are you planning on moving forward?
> 
> Hope everything's fine due to current situation 🤞

All is good, I plan to finish this off, but I can't commit to when, I'm on holiday this week, I'm interested in giving more thought to service provider / constructor middleware assignment as part of this work, I think it'd be good to keep the global definitions out of config, as a first step in potentially moving more things out and into a service provider discovery pattern.

mfn:
ping!

I released a new version now; wanted to wait for this, but not so long.

Still looking forward to this one!

mfn:
ping!

Are you still interested in moving forward?

I _might_ also be interesting wrapping this up (but haven't given this a deeper thought yet)

stevelacey:
I'll try to take a look within the next 30 days but if it doesn't happen by then you can safely assume this is up for grabs 👍

stevelacey:
@mfn okay this should be good to go, I didn't do anything more with global middlewares than make it slightly easier to override and customise the definition via a method, I think it's easiest to let users roll their own solution there – they can supply the `$middleware` attribute at any point in their inheritance hierarchy, or override `middleware()` and make it do something different.

radd2004:
Ahh! excited for this to get merged in!

crissi:
still missing something here? and can I help?

mfn:
Time for review 😄 

I want to give this one a thorough check to ensure it's really useful, i.e. I've private project which could make use of this and simplify the current approach and I want to take the time to validate the approach works.

crissi:
> Time for review 😄
> 
> I want to give this one a thorough check to ensure it's really useful, i.e. I've private project which could make use of this and simplify the current approach and I want to take the time to validate the approach works.

okay, 

already found a couple of use cases that i need at work. 

- changing params in a query from int to string, without making the queries fail for the consumers of the api
- rate limiting specific mutations.

alanwillms:
This PR is awesome, I was manually intercepting `getResolver()` method so I could do my magic there.

rt3norio:
Hello!
Just for record:

Today we once again needed this feature.

When calling the Graphql endpoint with wrong parameters, the request ended even before the resolvers that was instrumented to grab user/IP/etc so we can identify/log the request and notify the third party client.

any forecast on this going live?

illambo:
Hi, any news about this PR ? Thanks

websitevirtuoso:
Yes. Do we have any news? Great feature

mfn (approved):
This is a truly stellar PR, absolutely great job @stevelacey 

Apologies for the super long delay, but 2020 wasn't an easy year so far…

Additionally, although the changes are not very intrusive, they go into the heart of the graphql code and I made it a requirement for my review to run this code on a private project which already was using a middleware-alike code and I wanted to know how easy/possible the transformation is => and it was super-easy, basically worked out of the box.

I found some super small issues I could not push in this PR due to missing permissions, so I just added them in https://github.com/rebing/graphql-laravel/pull/699

I did not touch your commits so I hope that merging #699 will also mark this one as merged 🤞 

Again, sorry for the long delay and thanks for the great work! 👍

mfn:
It worked 😅, PR is properly marked as merged.

mfn:
For anyone interested => you can test this via https://github.com/rebing/graphql-laravel/releases/tag/6.1.0-rc1

stevelacey:
@mfn great news 🎉 🙌

PR #595, opened 2020-02-27
Standardize readme code blocks

stevelacey:


src/GraphQL.php
```diff
@@ -34,12 +34,14 @@ class GraphQL
 
     /** @var array<array|Schema> */
     protected $schemas = [];
+
```
mfn: I wish php-cs-fixer would do that, hate manual changes here :)

Issue #568, opened 2019-12-22
multiple schema on lumen not find query or mutation


amirhossein320:
**i receive this error**

 **Cannot query field \"users\" on type \"Query\".** 

**middleware in admin schema work  but query and mutation not work**

**my schemas**

```
'schemas' => [
        'default' => [
            'query' => [
            ],
            'mutation' => [
            ],
            'middleware' => [],
            'method' => ['get', 'post'],
        ],
        'admin' => [
            'query' => [
                'users' => UsersQuery::class,
            ],
            'mutation' => [
            ],
            'middleware' => ['admin_auth'],
            'method' => ['get', 'post'],
        ],
    ],
```
my **url** request on postman 

**http://localhost:8000/graphql/admin**

```
query fetchUsers{
    
    users(limit:20,  page: 2) {
        data{
            id, profile{
                name
            }
        },
        total,
        per_page

    }
}
```

mfn:
https://github.com/rebing/graphql-laravel/issues/509
https://github.com/rebing/graphql-laravel/issues/204

Contributions / fixes welcome!

LuizVisnadi:
Having the same problem on Lumen, I've looked into this package's code and found the possible source of error.

https://github.com/rebing/graphql-laravel/blob/20349a5aa0997b996b799147362e8343134e1420/src/GraphQLController.php#L31-L34

On line 32, changing the comparison from **"> 1"** to **"> 0"** seems to fix the problem.

Vespira:
I've experienced the same issue, and I think in this piece of code : 

```php
public function query(Request $request, string $schema = null): JsonResponse
    {
        $middleware = new GraphQLUploadMiddleware();
        $request = $middleware->processRequest($request);

        // If there are multiple route params we can expect that there
        // will be a schema name that has to be built
        $routeParameters = $this->getRouteParameters($request);
        if (count($routeParameters) > 1) {
            $schema = implode('/', $routeParameters);
        }
```

for some reason, Lumen does not fill the `$schema` arg, so when we have only one route depth, `$schema` is still null, hence only the **default_schema** work.
In case of error in the schema targeting, all queries and mutations are triggered on the default schema.

Anyway, @LuizVisnadi do you want to open a pull request to push this little change ?
This function works with only one route parameter, so I don't think it's impacting ? But mainteners have a wider understanding so, I hope someone could make us a feedback please :)

mfn:
[...]

PR #599, opened 2020-03-04
[LUMEN] Fix multiple schemas on query/mutation

Vespira:
## Summary
`$schema` arg received in query function is always `null` and if we have only one element in `$routeParameters`, `$schema` remain empty, then the query always hit default schema.

## Type of change
switching condition to fill `$schema`, from `if (count($routeParameters) > 1)` to `if (count($routeParameters) > 0)`

### Checklist
- [ ] Update the README.md

### Links
- Fixed https://github.com/rebing/graphql-laravel/issues/568

### DIFF
--- a/src/GraphQLController.php
+++ b/src/GraphQLController.php
@@ -29,7 +29,7 @@ class GraphQLController extends Controller
         // If there are multiple route params we can expect that there
         // will be a schema name that has to be built
         $routeParameters = $this->getRouteParameters($request);
-        if (count($routeParameters) > 1) {
+        if (count($routeParameters) > 0) {
             $schema = implode('/', $routeParameters);
         }
 


Vespira:
it appears to fail at dependency installation with composer, in this travis Job :  

> PHP: 7.3
> LARAVEL='^6.0' INSTALL_LARAVEL=1 INSTALL_LUMEN=1

But the fact is, I didn't change anything in composer.json or composer-lock file ... is it something that happens frequently ?

mfn:
> is it something that happens frequently ?

Hmm, not frequently but it can…
… still busy and some other PRs have more priority so pleas bear with me here; thx.

src/GraphQLController.php
```diff
@@ -29,7 +29,7 @@ public function query(Request $request, string $schema = null): JsonResponse
         // If there are multiple route params we can expect that there
         // will be a schema name that has to be built
         $routeParameters = $this->getRouteParameters($request);
-        if (count($routeParameters) > 1) {
+        if (count($routeParameters) > 0) {
```
mfn: I don't think the fix is correct here and should rather be within `\Rebing\GraphQL\GraphQLController::getRouteParameters`, can you investigate?

Vespira: > 
> 
> I don't think the fix is correct here and should rather be within `\Rebing\GraphQL\GraphQLController::getRouteParameters`, can you investigate?

Thanks mfn for having a look at my PR :) I don't see what can be fixed in the getRouteParameters as it already handle the Lumen case, and return the only one element in the route param available ..

mfn: Yes, so my impression is "if it's broken with Lumen, it need to be fixed in the Lumen code path".

This change here affects _all_ environments == also Laravel, so I'm not confident this is correct (as no one reported problems with Laravel).

Could I explain my position / concerns?

thanks

Vespira: Ok I get your point.

The situation here is kind of tricky because with Lumen and only one element in the route URL, the schema is never assigned, and the default schema will always be picked.

Could I write something like this instead : 

```php
$routeParameters = $this->getRouteParameters($request);
if (count($routeParameters) > Helpers::isLumen() ? 0 : 1) {
        $schema = implode('/', $routeParameters);
}
```
This fix wouldn't be necessary if we could pass a schema name directly into function arg : 
` public function query(Request $request, string $schema = null): JsonResponse {}`

mfn: > if (count($routeParameters) > Helpers::isLumen() ? 0 : 1) {

But you can to the same fix _within_ `getRouteParameters` in a lumen specific code path, no?

> This fix wouldn't be necessary if we could pass a schema name directly into function arg :

That would probably require adapting the `routes.php` => I'm open to that, but would see some concrete code.

Also we would definitely need a test for this.

Vespira: Sorry I'm on two parallel tasks, and I'm fairly new to PHP so I may be annoying. But my feeling here is that the `count($routeParameters) > 1` is not right, why would we want to take default schema, if there is one path element anyway ? it still seems valid.

Unfortunatly I don't know how it's supposed to behave with Laravel (what are contained in the `route()->parameters`; but my guess it that it sends the full path as an array in the $parameters value so for instance accessing the 'private' graphql schema would return : 
array('graphql', 'private') when Lumen is returning array('private'). So I guess we could just check if there is a 'graphql' element in the array, and remove it, and treat all values from getRouteParameters in the same way.

I'll update the PR to reflect that and try to update tests

Vespira: Ok I took some time again to look at the code, and I finally found something amazing 😲 

Into **graphql.php** config file, if you replace the academic `{graphql_schema?}` with `{schema}` in Lumen, it works so hard that I can't even think about how many hours I lost trying to wrap my head around it. I spotted the good place to log into RoutesRequests.php file from Lumen Framework, and I saw that the pattern given as a config key '**routes**' in graphql.php was used to insert a into the parameters the schema with this form (e.g private schema ) : 
`["private": "private"]`

Now if we change the config key to `{schema}` it will then automatically add into parameters : 
`["schema": "private"]`

Hence, the GraphQLController gets hit with correct schema value.

🗡 🗡 I could easily do Sepukku, if we were 600 years back in the past

Jokes aside, I will close this non-sense PR as it works with Lumen, BUT I will open another PR to update documentation about that.

Issue #602, opened 2020-03-13
Subquery args validations


illambo:
### Versions:
- graphql-laravel Version: 5.0.0-rc2
- Laravel Version: 6.18.1 
- PHP Version: 7.2.22

### Question:
How is it possible to validate the subquery args parameters?

Example:

```
class UserType extends GraphQLType
{
		// ...

		public function fields(): array
		{
			return [
				// ...

				// Relation
				'posts' => [
				'type'			=> Type::listOf(GraphQL::type('post')),
				'description'	=> 'A list of posts written by the user',
				'args'			=> [
					'date_from' => [
						'type'	=> Type::string(),
						'rules'	=> [ // this parameter has no effect
							'filled',
							'date'
						]
					],
				],
				'rules'			=> [ // this parameter has no effect
					'date_from' => [
						'filled',
						'date'
					]
				],
				// $args are the local arguments passed to the relation
				// $query is the relation builder object
				// $ctx is the GraphQL context (can be customized by overriding `\Rebing\GraphQL\GraphQLController::queryContext`
				'query'			=> function(array $args, $query, $ctx) {
					return $query->where('posts.created_at', '>', $args['date_from']);
				}
			]
		];
	}
}
```

Current solution (verbose):

```
class UserType extends GraphQLType
{

	// ...

	public function fields(): array
	{
		return [
			// ...

			// Relation
			'posts' => [
				'type'				=> Type::listOf(GraphQL::type('post')),
				'description'	=> 'A list of posts written by the user',
				'args'				=> [
					'date_from'	=> [
						'type' => Type::string(),
					],
				],
				// $args are the local arguments passed to the relation
				// $query is the relation builder object
				// $ctx is the GraphQL context (can be customized by overriding `\Rebing\GraphQL\GraphQLController::queryContext`
				'query'			=> function(array $args, $query, $ctx) {
					$validator = \Illuminate\Support\Facades\Validator::make($args, [
						'date_from' => [
							'filled',
							'date'
						]
					]);

					if ($validator->fails()) throw new \Rebing\GraphQL\Error\ValidationError('validation', $validator);
					
					return $query->where('posts.created_at', '>', $args['date_from']);
				}
			]
		];
	}
}
```

There is a better way?
Thanks

crissi:
confirmed that it does not seem to work. I will try take a deeper look

PR #608, opened 2020-03-21
validate field arguments

crissi:
## Summary
Fixes https://github.com/rebing/graphql-laravel/issues/602

Rules never worked in field arguments.

## Type of change

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ CHANGELOG
 2019-04-03, 5.0.0
 -----------------
 ### Added
+- Add support for validation in field arguments [\#608 / crissi](https://github.com/rebing/graphql-laravel/pull/608)
 - Support Laravel 7 [\#597 / exodusanto](https://github.com/rebing/graphql-laravel/pull/597)
 - Add support for custom authorization message [\#578 / Sh1d0w](https://github.com/rebing/graphql-laravel/pull/578)
 - Add support for macros on the GraphQL service/facade [\#592 / stevelacey](https://github.com/rebing/graphql-laravel/pull/592)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,12 @@ CHANGELOG
 
 [Next release](https://github.com/rebing/graphql-laravel/compare/5.0.0...master)
 --------------
+### Added
+- Add support for validation in field arguments [\#608 / crissi](https://github.com/rebing/graphql-laravel/pull/608)
 
 2019-04-03, 5.0.0
 -----------------
 ### Added
-- Add support for validation in field arguments [\#608 / crissi](https://github.com/rebing/graphql-laravel/pull/608)
 - Support Laravel 7 [\#597 / exodusanto](https://github.com/rebing/graphql-laravel/pull/597)
 - Add support for custom authorization message [\#578 / Sh1d0w](https://github.com/rebing/graphql-laravel/pull/578)
 - Add support for macros on the GraphQL service/facade [\#592 / stevelacey](https://github.com/rebing/graphql-laravel/pull/592)
--- a/src/Support/Field.php
+++ b/src/Support/Field.php
@@ -123,9 +123,6 @@ abstract class Field
             // 4 (!) - added by this library, encapsulates creating a `SelectFields` instance
             $arguments = func_get_args();
 
-            // Validate arguments in fields
-            $this->validateFieldArguments($arguments[3]);
-
             // Validate mutation arguments
             $args = $arguments[1];
 
@@ -134,9 +131,13 @@ abstract class Field
             if (count($rules)) {
                 $validator = $this->getValidator($args, $rules);
                 if ($validator->fails()) {
+
                     throw new ValidationError('validation', $validator);
                 }
             }
+
+            // Validate arguments in fields
+            $this->validateFieldArguments($arguments[3]);
 
             $arguments[1] = $this->getArgs($arguments);
 
--- a/src/Support/Field.php
+++ b/src/Support/Field.php
@@ -20,6 +20,13 @@ use ReflectionMethod;
  */
 abstract class Field
 {
+    /**
+     * The depth the SelectField and ResolveInfoFieldsAndArguments classes traverse
+     *
+     * @var integer
+     */
+    protected $depth = 5;
+
     protected $attributes = [];
 
     /**
@@ -86,12 +93,16 @@ abstract class Field
         return array_merge($argsRules, $rules);
     }
 
-    public function validateFieldArguments(ResolveInfo $resolveInfo): void
+    /**
+     * @param array<string,mixed> $fieldsAndArgumentsSelection
+     * @return void
+     */
+    public function validateFieldArguments(array $fieldsAndArgumentsSelection): void
     {
-        [$argumentValues, $argsRules] = (new RulesInFields($this->type(), $resolveInfo))->get();
 
+        $argsRules = (new RulesInFields($this->type(), $fieldsAndArgumentsSelection))->get();
         if (count($argsRules)) {
-            $validator = $this->getValidator($argumentValues, $argsRules);
+            $validator = $this->getValidator($fieldsAndArgumentsSelection, $argsRules);
             if ($validator->fails()) {
                 throw new ValidationError('validation', $validator);
             }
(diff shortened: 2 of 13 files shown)


crissi:
Depth is hardcoded, would adding a static setter be sufficient?

ResolveInfoFieldsAndArguments is fired twice if you are also using SelectField, maybe this could be moved, out, any suggestion to how you would do that @mfn ?

mfn:
@crissi ready for review?

crissi:
yes

src/Support/RulesInFields.php
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * @param  array<string,mixed>|string|callable  $rules
+     * @param  array<string,mixed>  $arguments
+     * @return array<string,mixed>|string
+     */
+    protected function resolveRules($rules, array $arguments)
+    {
+        if (is_callable($rules)) {
+            return call_user_func($rules, $arguments, $this->fieldsAndArguments);
```
mfn, line 59: What would be the use case to provide the full arguments too?

I see that closure and the test, but they don't make meaningful use of it

src/Support/RulesInFields.php
```diff
@@ -0,0 +1,105 @@
...
+
+            //If field doesn't exist on definition we don't select it
+            try {
+                if (method_exists($parentType, 'getField')) {
+                    $fieldObject = $parentType->getField($name);
+                } else {
+                    continue;
+                }
+            } catch (InvariantViolation $e) {
+                continue;
```
mfn, line 87: Currently no test covers this case

crissi: removing the catch, since it does not seem to ever get hit

src/Support/Field.php
```diff
@@ -123,6 +145,11 @@ protected function getResolver(): ?Closure
                 }
             }
 
+            $fieldsAndArguments = (new ResolveInfoFieldsAndArguments($arguments[3]))->getFieldsAndArgumentsSelection($this->depth);
```
mfn, line 148: No the big prize question: how can we avoid resolving this if there aren't going to be any validations (they're optional after all)? Do you see any way to achieve this?

Seems like a catch-22: only when we run `RulesInFields` for which we already have to extract "fields and arguments" we can then know that we wouldn't have to run it 😄

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ CHANGELOG
 --------------
 
 ### Added
+- Add support for validation in field arguments [\#608 / crissi](https://github.com/rebing/graphql-laravel/pull/608)
```
mfn, line 8: Since this code already moved, can you after rebase please make sure it's still under "Next release"? thanks!

src/Support/Field.php
```diff
@@ -165,11 +192,22 @@ protected function getResolver(): ?Closure
...
+     * @return SelectFields
+     */
+    private function instanciateSelectFields(array $arguments, int $depth = null, array $fieldsAndArguments): SelectFields
     {
         $ctx = $arguments[2] ?? null;
 
-        return new SelectFields($arguments[3], $this->type(), $arguments[1], $depth ?? 5, $ctx);
+        if ($depth !== null && $depth !== $this->depth) {
+            $fieldsAndArguments = (new ResolveInfoFieldsAndArguments($arguments[3]))
+                ->getFieldsAndArgumentsSelection($depth);
```
mfn, line 207: Nice catch!

src/Support/Field.php
```diff
@@ -165,11 +192,22 @@ protected function getResolver(): ?Closure
...
     }
 
-    private function instanciateSelectFields(array $arguments, int $depth = null): SelectFields
+    /**
+     * @param array<int,mixed> $arguments
+     * @param int $depth
+     * @param array<string,mixed> $fieldsAndArguments
+     * @return SelectFields
+     */
+    private function instanciateSelectFields(array $arguments, int $depth = null, array $fieldsAndArguments): SelectFields
```
mfn, line 201: I would move around the args, feels strange to have an optional parameter right in the middle (i.e. I suggest to make it the last one)

src/Support/RulesInFields.php
```diff
@@ -0,0 +1,93 @@
...
+
+class RulesInFields
+{
+    /**
+     * @var  Type
+     */
+    protected $parentType;
+
+    /**
+     * @var array<mixed,mixed>
```
mfn, line 18: I think this is still `string,mixed`, no?

src/Support/RulesInFields.php
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * @param  array<string,mixed>|string|callable  $rules
+     * @param  array<string,mixed>  $arguments
+     * @return array<string,mixed>|string
+     */
+    protected function resolveRules($rules, array $arguments)
+    {
+        if (is_callable($rules)) {
+            return call_user_func($rules, $arguments);
```
mfn, line 50: Maybe just … ?
```suggestion
            return $rules($arguments);
```

src/Support/RulesInFields.php
```diff
@@ -0,0 +1,93 @@
...
+
+        foreach ($fields as $name => $field) {
+            $key = $prefix === null ? $name : "{$prefix}.{$name}";
+
+            //If field doesn't exist on definition we don't select it
+            if (method_exists($parentType, 'getField')) {
+                $fieldObject = $parentType->getField($name);
+            } else {
+                continue;
+            }
```
mfn, line 76: Suggest to de-nest this _a bit_; this changed logic would also better match the comment:
```suggestion
            //If field doesn't exist on definition we don't select it
            if (!method_exists($parentType, 'getField')) {
                continue;
            }
            
            $fieldObject = $parentType->getField($name);
```

src/Support/SelectFields.php
```diff
@@ -32,32 +31,36 @@ class SelectFields
...
-        $requestedFields = $this->getFieldSelection($info, $queryArgs, $depth);
+        $requestedFields = $this->getFieldSelection($fieldsAndArguments, $queryArgs);
         $fields = self::getSelectableFieldsAndRelations($queryArgs, $requestedFields, $parentType, null, true, $ctx);
         $this->select = $fields[0];
         $this->relations = $fields[1];
     }
 
-    private function getFieldSelection(ResolveInfo $resolveInfo, array $args, int $depth): array
+    /**
+     * Undocumented function.
```
mfn, line 53: lol 😄 

I would vote for completely removing the method and just construct the `$requestedFields` in the constructor. It almost does nothing, it's private so no one can customize this anyway, i.e. see no worth in keeping.

crissi: oops:-)

src/Support/Field.php
crissi: I could probably write a function that checks if any arguments exists that would be less costly than the ResolveInfoFieldsAndArguments, and then only run ResolveInfoFieldsAndArguments if using selectfields and arguments... but not sure how much it will actually give.

mfn: You're right, let's scrap that for now!

mfn:
The job with the integration test is failing, I'll restart it

mfn:
hm, nope, fails too https://travis-ci.org/github/rebing/graphql-laravel/jobs/677399766#L380 :
```
PHP Fatal error:  Allowed memory size of 1610612736 bytes exhausted (tried to allocate 4096 bytes) in phar:///home/travis/.phpenv/versions/7.3.17/bin/composer/src/Composer/DependencyResolver/RuleWatchGraph.php on line 52
```
That's 1.6GB so no point in bumping the memory_limit, something is off here.

My first idea maybe something in composer changed, a newer version, but https://travis-ci.org/github/rebing/graphql-laravel/jobs/677399766#L148
> `Composer version 1.8.4 2019-02-11 10:52:10`

I pushed a commit to see how it goes when we `composer self-update`…

mfn:
https://travis-ci.org/github/rebing/graphql-laravel/jobs/677442929#L152
```
Composer version 1.10.5 2020-04-10 11:44:22

$ composer self-update
You are already using composer version 1.10.5 (stable channel).
```
Ach #fail

Meh, must be something else as the trace comes from within the resolver…

I'll try the composer 2.0 preview, maybe we're lucky :)

mfn:
> I'll try the composer 2.0 preview, maybe we're lucky :)

We were 🍀 !

mfn:
@crissi you might already noticed, this added a regression see https://github.com/rebing/graphql-laravel/issues/627 

Please feel free to "revert my revert" and apply a fix (probably the InvariantException thingy, 95% sure)

Thanks!

(and sorry for my feedback: pretty sure I was the reason why we removed it)

crissi:
Np, will make a fix later

Sendt: Thursday, April 23, 2020 11:38:11 AM
Cc: Christian <c2n@live.dk>; Mention <mention@noreply.github.com>
Emne: Re: [rebing/graphql-laravel] validate field arguments (#608)


@crissi<https://github.com/crissi> you might already noticed, this added a regression see #627<https://github.com/rebing/graphql-laravel/issues/627>

Please feel free to "revert my revert" and apply a fix (probably the InvariantException thingy, 95% sure)

Thanks!

(and sorry for my feedback: pretty sure I was the reason why we removed it)

PR #612, opened 2020-03-29
fix alias relationships not firing its own query

### LINKED ISSUE #604: SelectFields will only fire relationship query once and therefore with wrong arguments
I was looking into this https://github.com/rebing/graphql-laravel/issues/602 and fell over this bug.

With this query:
```
query ($proId: Int, $proId2: Int) {
  test2 {
    name

    alias1: profile(profileId: $proId2) {
      contentType
    }
    
    alias2:profile(profileId: $proId) {
      contentType
    }

  }
}
```
```
{
  "proId": 2,
  "proId2": 4,
  "id": 70
}
```

image

The profile relationship is only run once and therefore the different arguments given to the custom query clause is not taken into account, only the last query get the right args.

It seems a bit to hard to fix since it is not easily possible to alias eloquent relationships.

Unfortunately something like this is not possible. This would have made it a simplier fix
`User::with('relationship', 'relationship AS other')->get();`

crissi:
## Summary
fixes #604 

## Type of change

### Checklist

### DIFF
--- a/src/Support/AliasedRelationships/AddModelRelationshipScope.php
+++ b/src/Support/AliasedRelationships/AddModelRelationshipScope.php
@@ -1,0 +1,35 @@
+<?php
+namespace Rebing\GraphQL\Support\AliasedRelationships;
+
+use Illuminate\Database\Eloquent\Scope;
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Model;
+
+class AddModelRelationshipScope implements Scope
+{    
+    /**
+     * @var array<string,string>
+     */
+    private $relationships;
+
+    /**
+     * @param array<string,string> $relationships
+     */
+    public function __construct(array $relationships = [])
+    {
+        $this->relationships = $relationships;
+    }
+
+    public function extend(Builder $builder): void {
+        foreach($this->relationships as [$graphqlAlias, $relationship]) {
+            $builder->macro($graphqlAlias, function (Builder $builder) use($relationship) {
+                return $builder->getModel()->{$relationship}();
+            });
+        }
+    }
+
+    public function apply(Builder $builder, Model $model) {
+    
+    }
+    
+}
--- a/src/Support/AliasedRelationships/AddModelRelationshipScope.php
+++ b/src/Support/AliasedRelationships/AddModelRelationshipScope.php
@@ -1,12 +1,15 @@
 <?php
+
+declare(strict_types=1);
+
 namespace Rebing\GraphQL\Support\AliasedRelationships;
 
-use Illuminate\Database\Eloquent\Scope;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Scope;
 
 class AddModelRelationshipScope implements Scope
-{    
+{
     /**
      * @var array<string,string>
      */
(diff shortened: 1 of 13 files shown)


src/Support/AliasedRelationships/ModelRelationshipAdder.php
```diff
@@ -0,0 +1,20 @@
...
+class ModelRelationshipAdder
+{
+    /**
+     * @param class-string<Model> $model
+     * @param array<string,string> $relationships
+     * @return void
+     */
+    public static function add(string $model, array $relationships): void
+    {
+        $model::addGlobalScope(new AddModelRelationshipScope($relationships));
```
mfn, line 18: 👀 

Isn't that somehow dangerous (side-effects), permanently mutating the the models global scopes?

src/Support/AliasedRelationships/AddModelRelationshipScope.php
```diff
@@ -0,0 +1,38 @@
...
+     */
+    public function __construct(array $relationships = [])
+    {
+        $this->relationships = $relationships;
+    }
+
+    public function extend(Builder $builder): void
+    {
+        foreach ($this->relationships as [$graphqlAlias, $relationship]) {
+            $builder->macro($graphqlAlias, function (Builder $builder) use ($relationship) {
```
mfn, line 29: 👀

src/Support/SelectFields.php
```diff
@@ -28,7 +32,10 @@ class SelectFields
     private $select = [];
     /** @var array */
     private $relations = [];
-
+    /**
+     * @var array<class-string<Model>,array<string,string>>
+     */
+    private static $modelRelationshipsForAlias = [];
```
mfn, line 38: Wish this wouldn't be static.

I understand from the code it's cleaned up, still…

src/Support/AliasedRelationships/ModelRelationshipAdder.php
crissi: > I don't feel confident doing a good review on this one.
> 
> This shows the absurd complexity we've to put upon us because of the concept of the `SelectFields` and once more shows me this approach goes into the wrong direction somehow…
> 
> Leads me to believe `SelectFields` should be a standalone component and not in this library TBH 😄
> 
> Mutating models scopes, adding macros to builders .. this somehow looks dangerous to me 🤷‍♀️
> 
> When I think about I would implement a dedicated PHP server with this library (i.e. never restart, a daemon) I can foresee thus causing problems. Is there no other way to solve this aliasing / double-query problem?
> 
> Not really sure how to move forward with this one TBH.
> 
> > Leads me to believe `SelectFields` should be a standalone component and not in this library TBH 😄
> 
> To clarify this, I believe the proper approach is to use a dataloader concept to solve these things in general (that's what I'm doing in my projects, I'm not using SelectFields)

I love the SelectFields and use it for 99% of my queries, makes it so easy to set up

crissi: > 👀
> 
> Isn't that somehow dangerous (side-effects), permanently mutating the the models global scopes?

yes it is definitely a side effect, but since SelectFields are made to be used with eloquent's ->with(). The only way to fix this is to generate the relationships.

The side effect is minimized by generating relationships with **_generated_** in front of all generated relationships, so I would say this makes it as safe as possible.

src/Support/SelectFields.php
crissi: yeah me too

src/Support/AliasedRelationships/ModelRelationshipAdder.php
crissi: yeah we might need to tore the relationship down if some one uses the same relationship again

mfn:
@crissi not sure if/how you want to move forward here.

Besides all the complexity, for me the current no-go definitely are the side-effects of mutating the models behaviour (and also the macro thing); I don't think this is acceptable to be done from such a library.

crissi:
I see you point, but people would expect aliases to work since they are a feature in graphql.

could we make it an opt in feature in the config?

I want to tighten it up, after some of the cases you mention in the comments.

mfn:
Sorry to sound stubborn, but as long as the code mutates the models I don't want to accept this (config or not; and I think it's a bad idea to use a "config switch" for that).

crissi:
> Sorry to sound stubborn, but as long as the code mutates the models I don't want to accept this (config or not; and I think it's a bad idea to use a "config switch" for that).

fair

PR #614, opened 2020-03-30
[LUMEN] Update documentation

Vespira:
## Summary
Updated README.md to add a part in the Lumen part, where we need to change the routes key in graphql.php conf file in order to receive correctly this parameter (tested with Lumen > 5.8)

_Continuation from https://github.com/rebing/graphql-laravel/pull/599_

## Type of change

### Checklist
- [x] Update the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,9 @@ To work this around:
 - Point to your local GraphiQL view: change `graphql.view` to `'vendor/graphql/graphiql'`
 - Modify your file `resources/views/vendor/graphql/graphiql.php` and remove the call
 
+In the `graphql.php` file, you may want to change the 'routes' key value from `{graphql_schema?}` to `{schema}`.
+By doing this, you will pass the correct schema value into the GraphQLController.
+
 ## Usage
 
 - [Laravel GraphQL](#laravel-graphql)
@@ -148,7 +151,7 @@ Schemas are required for defining GraphQL endpoints. You can define multiple sch
 in addition to the global middleware. For example:
 
 ```php
-'schema' => 'default_schema',
+'default_schema' => 'default',
 
 'schemas' => [
     'default' => [


README.md
```diff
@@ -98,6 +98,9 @@ To work this around:
 - Point to your local GraphiQL view: change `graphql.view` to `'vendor/graphql/graphiql'`
 - Modify your file `resources/views/vendor/graphql/graphiql.php` and remove the call
 
+In the `graphql.php` file, you may want to change the 'routes' key value from `{graphql_schema?}` to `{schema}`.
```
mfn, line 101: Interesting!

I don't use Lumen myself, so it's hard for me to properly verify this.
I though wonder if, weapon with this knowledge, why can't provide a proper programmatic solution?

> I don't use Lumen myself, so it's hard for me to properly verify this.

However last year I set up `tests/integration-lumen.sh` so we _have some lumen_ testing.

I think it would be great if could show this problem with the integration test and based on that provide a fix

Vespira: hi @mfn it's been a while. I just remembered of this subject and forgot to provide details, I'll try to do so.
In your lumen testing, in the multi-level schema, you are setting as default schema the ``multi/level``, so the query work. However, in my case, my default schema is "auth" and if I query it with the default configuration file, it was ok. But as soon as I tried to query other schemas, it wasn't working (outside the "default" or "default_schema" key config)

mfn: I'm open adapting/changing all of this, but hesitant without a test!

I think it would be highly appreciated if you can take a look at the lumen integration test and come up with a scenario for that.

mfn:
I'm rejecting the PR for the time being, as I rather would figure out a proper programmatic solution than having ppl to steer to the README for this (or drop Lumen support alltogether).

PR #622, opened 2020-04-11
Input type required_with prefix

stevelacey:
## Summary

Input type validation rules need to prefix the field reference in the `required_with` rules, otherwise the rules would have to be pushed up to the query level.

## Type of change

### Checklist

### DIFF
--- a/src/Support/Rules.php
+++ b/src/Support/Rules.php
@@ -44,10 +44,32 @@ class Rules
      * @param  array<string,mixed>  $arguments
      * @return array<string,mixed>|string
      */
-    protected function resolveRules($rules, array $arguments)
+    protected function resolveRules($rules, ?string $prefix, array $arguments)
     {
         if (is_callable($rules)) {
             return call_user_func($rules, $arguments, $this->requestArguments);
+        }
+
+        if ($prefix && is_array($rules)) {
+            $rules = array_map(
+                function ($rule) use ($prefix) {
+                    if (is_string($rule) && preg_match('/^(required_with(?:out)?(?:_all)?):(.*)$/', $rule, $matches)) {
+                        [$rule, $name, $parameters] = $matches;
+
+                        $parameters = implode(',', array_map(
+                            function ($param) use ($prefix) {
+                                return "{$prefix}.{$param}";
+                            },
+                            explode(',', $parameters)
+                        ));
+
+                        $rule = "{$name}:{$parameters}";
+                    }
+
+                    return $rule;
+                },
+                $rules
+            );
         }
 
         return $rules;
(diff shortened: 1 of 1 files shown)


stevelacey:
@mfn what do you think about this, shall I finish it up? I figure this needs to work

mfn:
Sorry for missing this. Was pretty sure I already looked at it.

The manual preg_match for a rulename feels strange but I undertsand, there's no other way to make this probably work.

Would like to tie in @crissi who did some great work on validation what he things?

crissi:
> @mfn what do you think about this, shall I finish it up? I figure this needs to work
> 
> **Edit:** I actually have some other rules that I've written that behave like `required_with` and accept a field arg that would continue to not work here, I wonder if there's a more comprehensive solution to the problem; is there a good reason we're building out indexed rules for arrays as opposed a asterisk `x.*.y` rule? I assume this is important and necessary, but maybe it's not. Perhaps validation needs a rework.

 I remember this coming up during some work with rules, but I right now I actually can't see why it wouldn't work with *

crissi:
never worked with the required_with so not much to say here, but it makes sense that we need params from the same depth here.

The PR could use some tests though

src/Support/Rules.php
```diff
@@ -44,12 +44,34 @@ public function get(): array
...
+    protected function resolveRules($rules, ?string $prefix, array $arguments)
     {
         if (is_callable($rules)) {
             return call_user_func($rules, $arguments, $this->requestArguments);
         }
 
+        if ($prefix && is_array($rules)) {
+            $rules = array_map(
+                function ($rule) use ($prefix) {
+                    if (is_string($rule) && preg_match('/^(required_with(?:out)?(?:_all)?):(.*)$/', $rule, $matches)) {
```
mfn, line 56: Hm, don't we also support the array forms of rules, e.g. `['required_with:…', '…']` or am I on the completely wrong track here?

Yeah, I think tests to play around would help :)

stevelacey:
I somewhat lost appetite for this approach as soon as I realised that it wouldn't cover my custom `absent_with` extension (the inverse of `required_with`) without additional effort, to do that I'd also need to add a way to tell the package which rule parameters are "field parameters" and should thus be prefixed inside input types... that... or do something like "check if there's a field that === this name?" which might be a sketchy approach.

src/Support/Rules.php
stevelacey: Yeah `['required_with' => ['thing']]` might be a thing

stevelacey: Also this is apparently a thing

```php
                'rules' => function ($rules) {
                    return ['required'];
                },
```

Not sure how viable this approach is.

mfn:
> I somewhat lost appetite for this approach as

Shall we close the PR then 🤔?

stevelacey:
Yep, I went with query-level array rules (`thing.*.thing`) in the end, it's awkward but saved addressing this problem.

PR #637, opened 2020-06-05
Freeze the time to trigger date test

kylekatarnls:
## Summary
Solves https://travis-ci.org/github/rebing/graphql-laravel/jobs/693126257#L579

By freezing the time, you can actually rely on the predictable interval to be calculated (it won't depends on the time the test takes to run, the current hour, etc.)

Then `Carbon::now()->startOfDay()` is `Carbon::today()`

## Type of change

- Unit test fix

### Checklist
- [ ] Add a CHANGELOG.md entry: not needed: no impact on the package
- [ ] Update the README.md: NA

### DIFF
--- a/tests/Unit/InstantiableTypesTest/InstantiableTypesTest.php
+++ b/tests/Unit/InstantiableTypesTest/InstantiableTypesTest.php
@@ -9,8 +9,10 @@ use Rebing\GraphQL\Tests\TestCase;
 
 class InstantiableTypesTest extends TestCase
 {
-    public function testSomething(): void
+    public function testDateFunctions(): void
     {
+        Carbon::setTestNow('2020-06-05 12:34:56');
+
         $query = <<<'GRAQPHQL'
 {
     user {
@@ -24,8 +26,8 @@ GRAQPHQL;
 
         $result = $this->graphql($query);
 
-        $dateOfBirth = Carbon::now()->addMonth()->startOfDay();
-        $createdAt = Carbon::now()->startOfDay();
+        $dateOfBirth = Carbon::today()->addMonth();
+        $createdAt = Carbon::today();
 
         $expectedResult = [
             'data' => [


tests/Unit/InstantiableTypesTest/InstantiableTypesTest.php
```diff
@@ -9,8 +9,10 @@
 
 class InstantiableTypesTest extends TestCase
 {
-    public function testSomething(): void
+    public function testDateFunctions(): void
```
kylekatarnls, line 12: I guess `testSomething` was not meant to be the definitive name of the test.

tests/Unit/InstantiableTypesTest/InstantiableTypesTest.php
```diff
@@ -9,8 +9,10 @@
...
     {
+        Carbon::setTestNow('2020-06-05 12:34:56');
```
mfn, line 14: What I usually prefer to do is
```php
        Carbon::setTestNow(Carbon::now());
```
This makes it clear that the _actual test date_ given does not signify any relevance.

(I've seen tests with specific dates and then also had them literally in the tests with their offsets and such, quite annoying)

What do you think?

kylekatarnls:
Actually I just updated my rebing/graphql-laravel package and saw your Travis status was red. :)

tests/Unit/InstantiableTypesTest/InstantiableTypesTest.php
kylekatarnls: In your case, the test will fail if it's executed at 00:00:00, because at this moment now = today so the month (made by addMonth) will be full and you'll get "1 month" instead of "4 weeks".

But you can do:
```php
Carbon::setTestNow('12:34:56');
```

So it mocks time only and not the date part.

About doing `Carbon::setTestNow(Carbon::now())` it won't trigger the 29 February bugs, DST bugs, midnight shift, etc. until your tests are luckily ran on those particular moments. So I would say it does not worth it. You get better luck by looping over hundred random dates and times. There you actually has chances to fall on the particular cases you may want to detect.

Else, as you more likely test *your* package and not the PHP `DateTime` behavior or the `Carbon` behavior, then I would say mocking the date is as relevant as mocking your JSON input, user input etc. It's just an input you get like the other ones.

kylekatarnls: An other example `Carbon::now()->subYear()->diffForHumans()`  will return `1 year ago` unless you run on a 29th February. In this case, you get `11 months ago`. Because unless using `subYearNoOverflow`, we'll rely on the default behavior of `DateTime` and it will switch 29th February with the 1st March if it did not fall on a leap year.

kylekatarnls: You also may be interested in https://github.com/kylekatarnls/carbonite

The principle is: if you would have "Sunny" as a possible input in your tests, you would not try get the current weather, you would just try with 3 or 4 inputs "Cloudy", "Sunny", "Rain", "Snow", etc. and you would try to cover edge-cases of the weather API plugged in.

That's the same for date, it's part of the story you tell in your tests, and in tests you should prefer static inputs over calculated inputs, as you don't test the input, you test how your functions transform those inputs into the output you were expecting.

PR #648, opened 2020-06-30
Shorten SelectFields constructor using array unpacking

kylekatarnls:
## Summary
Syntax proposal: usage of array unpacking

Benefits:
- Avoid intermediate variable
- Avoid magic numbers 0 and 1
- Avoid ambiguous 'fields' name: fields[0] (which make think the first field only goes to select and the second one to relations while it's actually 2 groups of different things that have been retrieved and the unit is not the field)
- Solve a PHPStan error

## Type of change
Misc. change (syntax)


mfn:
Thank you @kylekatarnls !

PR #658, opened 2020-07-09
Schema types fix

stevelacey:
## Summary

Specifying additional types per schema is broken, this fixes that and updates the tests to exercise the problem.

## Type of change

### Checklist
- [x] Tests updated

### DIFF
--- a/src/GraphQL.php
+++ b/src/GraphQL.php
@@ -71,6 +71,9 @@ class GraphQL
         $schemaQuery = $schema['query'] ?? [];
         $schemaMutation = $schema['mutation'] ?? [];
         $schemaSubscription = $schema['subscription'] ?? [];
+        $schemaTypes = $schema['types'] ?? [];
+
+        $this->addTypes($schemaTypes);
 
         $query = $this->objectType($schemaQuery, [
             'name' => 'Query',
@@ -88,21 +91,11 @@ class GraphQL
             'query' => $query,
             'mutation' => ! empty($schemaMutation) ? $mutation : null,
             'subscription' => ! empty($schemaSubscription) ? $subscription : null,
-            'types' => function () use ($schema) {
+            'types' => function () {
                 $types = [];
-                $schemaTypes = $schema['types'] ?? [];
 
-                if ($schemaTypes) {
-                    foreach ($schemaTypes as $name => $type) {
-                        $opts = is_numeric($name) ? [] : ['name' => $name];
-                        $objectType = $this->objectType($type, $opts);
-                        $this->typesInstances[$name] = $objectType;
-                        $types[] = $objectType;
-                    }
-                } else {
-                    foreach ($this->getTypes() as $name => $type) {
-                        $types[] = $this->type($name);
-                    }
+                foreach ($this->getTypes() as $name => $type) {
+                    $types[] = $this->type($name);
                 }
 
                 return $types;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ CHANGELOG
 [Next release](https://github.com/rebing/graphql-laravel/compare/6.3.0...master)
 --------------
 
+### Added
+- Support for per-schema types [\#658 / stevelacey](https://github.com/rebing/graphql-laravel/pull/658)
+
 2021-03-12, 6.3.0
 -----------------
 
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ Schemas are required for defining GraphQL endpoints. You can define multiple sch
 in addition to the global middleware. For example:
 
 ```php
-'schema' => 'default_schema',
+'schema' => 'default',
 
 'schemas' => [
     'default' => [
@@ -167,6 +167,9 @@ in addition to the global middleware. For example:
         'mutation' => [
             ExampleMutation::class,
         ],
+        'types' => [
+        ],
+]
     ],
     'user' => [
         'query' => [
@@ -174,6 +177,8 @@ in addition to the global middleware. For example:
         ],
         'mutation' => [
 
+        ],
+        'types' => [
         ],
         'middleware' => ['auth'],
     ],
@@ -207,7 +212,9 @@ class DefaultSchema implements ConfigConvertible
             ],
             'mutation' => [
                 ExampleMutation::class,
-            ]
+            ],
+            'types' => [
+            ],
         ]
     }
 }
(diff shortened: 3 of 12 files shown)


mfn:
That rips out the heart of the lazy type loader (see https://github.com/rebing/graphql-laravel/pull/405 ) and tests are breaking.

See also the discussion in https://github.com/rebing/graphql-laravel/issues/390#issuecomment-513495786 regarding the performance improvements

It's quite possible it works in your project (and probably would also in mine), but we wouldn't reap the benefits of lazy loading on larger schemas when not all types are necessary for a given request (think about a schema with 100 types and 100 queries but any given request only requires a fraction of them).

stevelacey:
@mfn any suggestions for how to get it to do both?

stevelacey:
@mfn how about this? Seems to work and the tests pass.

mfn:
Before taking a deeper look I decided to re-test the benchmarks from https://github.com/rebing/graphql-laravel/issues/390#issuecomment-513494782 and https://github.com/rebing/graphql-laravel/issues/390#issuecomment-513495786

I can report back that, although the absolute numbers changed, the relative improvement is also still there 👍 

----
For this to work I only applied the diff from this PR manually as this uses an older version and I wasn't keen yet do upgrade this.

- `lazyload_types=false` => ~730ms
- `lazyload_types=true` => ~530ms

I couldn't see any difference between the existing lazyload_types and from this PR 👍 

----
@stevelacey since this PR supposedly fixes something 😏 can you please add a test showing this? A test I expect would not work without this change. Thanks!

stevelacey:
@mfn awesome, good to know. I've updated the tests, they now fail on the previous commit with `CustomExample type not found`, which is what is fixed here 🤘

src/GraphQL.php
```diff
@@ -70,6 +70,9 @@ public function schema($schema = null): Schema
         $schemaQuery = $schema['query'] ?? [];
         $schemaMutation = $schema['mutation'] ?? [];
         $schemaSubscription = $schema['subscription'] ?? [];
+        $schemaTypes = $schema['types'] ?? [];
+
+        $this->addTypes($schemaTypes);
```
mfn, line 75: This one is interesting.

Every time a query is now run, it adds the schema types to the internal `types` list.

Previously this was only done once when constructing _the_ GraphQL instance.

Multiple invocation within the same process of the same query will work correctly, as `types` are stored by their name.

Now, if multiple schemas with same named types are used... they will overwrite previous one but probably "just" work. I guess.

It's in fact similar to `typesInstances` with the difference that they're **cleared** before every query, so instances of different schemas are never mixed.

I still have to think about the impact here but if you've any insights here, please share them!

mfn:
Ah, one more thing I noticed: the readme from the beginning did give examples of multiple schemas, i.e. starts off with the `'default'` schema etc.

So the best practice for defining queries/mutation is in fact per schema.

Interestingly, for the types it was never. We're basically promoting global types.

That's also emphasized by the default config we publish: there is even a `types` key there to begin with!

I wonder if this general sentiment should be changed, to focus on the whole "per schema" experience and only use global types for specific cases (something like the `UploadType` or probably any custom scalar types).

Any feeling about this?

src/GraphQL.php
stevelacey: @mfn the thing is this whole mechanism seems to rely on php's request lifetime already, and I've not addressed that problem to reach my goal here – really we should do better, but it looks like a significant piece of work to get there I think. The `GraphQL::type(...)` helper would need to become schema-aware, and know to look within a separate pool of independently built/cached types depending on the current schema in use.

I figured the tests would exercise this problem and break for this reason, but they didn't, probably because even when >1 queries are executed within the same test, they're happening against the same schema, the same as in a request. Since a request can only be performed against one schema at a time, I figured this solution would probably do for now, but really I'd prefer this scenario was handled better.

stevelacey:
> Ah, one more thing I noticed: the readme from the beginning did give examples of multiple schemas, i.e. starts off with the `'default'` schema etc.
> 
> So the best practice for defining queries/mutation is in fact per schema.
> 
> Interestingly, for the types it was never. We're basically promoting global types.
> 
> That's also emphasized by the default config we publish: there is even a `types` key there to begin with!
> 
> I wonder if this general sentiment should be changed, to focus on the whole "per schema" experience and only use global types for specific cases (something like the `UploadType` or probably any custom scalar types).
> 
> Any feeling about this?

Yeah, I agree, globally defined types shouldn't be the recommendation, in the project I've been working on it's resulted in 8 schema all with types half of them don't use. The most common reason for even having multiple schema is probably to differentiate public and private APIs, which likely have significant differences in their type offerings. In my case, only 1 of these 8 APIs is a public API, and it shouldn't receive the types used exclusively by my private APIs, nor populate the new schema I'm adding with types that I've deprecated.

Another thing worth noting is that sharing types between schema isn't hard even without a global types pool, encouraging users to define everything per schema and deprecate the feature entirely might be the way to go, it'll probably raise the quality of GraphQL APIs produced with this library – especially when you consider the tools like Graphidocs and GraphQL Playground that build documentation out of these type definitions.

mfn:
@crissi in case you're still around 😅 since you're the original author of the improved type loader, do you think you've time for a review?

My concern from https://github.com/rebing/graphql-laravel/pull/658#discussion_r453306205 isn't addressed yet, but I also found no good solution ATM.

The PR isn't yet finished in terms of tests, but I tested it successfully on a private project with per schema-types and also per-schema types mixed with global types etc. and everything worked out.

crissi:
> @crissi in case you're still around 😅 since you're the original author of the improved type loader, do you think you've time for a review?
> 
> My concern from [#658 (comment)](https://github.com/rebing/graphql-laravel/pull/658#discussion_r453306205) isn't addressed yet, but I also found no good solution ATM.
> 
> The PR isn't yet finished in terms of tests, but I tested it successfully on a private project with per schema-types and also per-schema types mixed with global types etc. and everything worked out.

Still around:) I can try to test the branch tomorrow with some production code and see if it breaks anything or slows execution down significantly

mfn:
@crissi that would be great!

> The PR isn't yet finished in terms of tests,

PS: done from my side!

mfn (approved):

mfn:
@crissi thanks, same experience here!

I'll probably maybe merge and release this later today.

mfn:
Almost forgot: @stevelacey thanks for the PR!

stevelacey:
Nice to see this finally merge! @bradez

Issue #640, opened 2020-06-10
[LUMEN] ErrorException on declaring class alias


LuizVisnadi:
Hello,

Currently i'm having a small issue with this package under Lumen. When i launch my unit tests (with phpunit), the class GraphQLLumenServiceProvider throws a ErrorException because it tries to assign an alias to 'Illuminate\Routing\Controller' but it already exists.

https://github.com/rebing/graphql-laravel/blob/1606e9a8fa751f449d818a39fd74182c5cb669f5/src/GraphQLLumenServiceProvider.php#L23-25

When adding `&& !class_exists('Illuminate\Routing\Controller')` on the condition line 23. It seems to correct the issue.

mfn:
Fixed via https://github.com/rebing/graphql-laravel/pull/508

PR #664, opened 2020-08-14
Fix problem with redeclaring the class in phpunit

ulrichdahl:
Suggested change from issue #640 works well.
https://github.com/rebing/graphql-laravel/issues/640

## Summary
When running phpunit tests you will get this error:
```
ErrorException: Cannot declare class Illuminate\Routing\Controller, because the name is already in use

/var/www/vendor/rebing/graphql-laravel/src/GraphQLLumenServiceProvider.php:24
/var/www/vendor/laravel/lumen-framework/src/Application.php:233
/var/www/bootstrap/app.php:100
/var/www/tests/TestCase.php:17
/var/www/vendor/laravel/lumen-framework/src/Testing/TestCase.php:55
/var/www/vendor/laravel/lumen-framework/src/Testing/TestCase.php:72
```
## Type of change
Simply check if the class already exists before creating the alias.

### Checklist
- [ ] Update the README.md

### DIFF
--- a/src/GraphQLController.php
+++ b/src/GraphQLController.php
@@ -9,9 +9,8 @@ use Illuminate\Container\Container;
 use Illuminate\Contracts\View\View;
 use Illuminate\Http\JsonResponse;
 use Illuminate\Http\Request;
-use Illuminate\Routing\Controller;
 
-class GraphQLController extends Controller
+class GraphQLController
 {
     /** @var Container */
     protected $app;
--- a/src/GraphQLLumenServiceProvider.php
+++ b/src/GraphQLLumenServiceProvider.php
@@ -20,7 +20,7 @@ class GraphQLLumenServiceProvider extends GraphQLServiceProvider
 
     public function register()
     {
-        if (class_exists('Laravel\Lumen\Routing\Controller')) {
+        if (class_exists('Laravel\Lumen\Routing\Controller') && !class_exists('Illuminate\Routing\Controller')) {
             class_alias('Laravel\Lumen\Routing\Controller', 'Illuminate\Routing\Controller');
         }
 
--- a/src/GraphQLLumenServiceProvider.php
+++ b/src/GraphQLLumenServiceProvider.php
@@ -6,6 +6,10 @@ namespace Rebing\GraphQL;
 
 use Rebing\GraphQL\Console\PublishCommand;
 
+if (class_exists('Laravel\Lumen\Routing\Controller')) {
+    class_alias('Laravel\Lumen\Routing\Controller', 'Illuminate\Routing\Controller');
+}
+
 class GraphQLLumenServiceProvider extends GraphQLServiceProvider
 {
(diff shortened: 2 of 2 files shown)


mfn:
Please also see the discussion in https://github.com/rebing/graphql-laravel/pull/508

ulrichdahl:
@mfn Problem found, the register method has to be idempotent within the same process execution to work with phpunit. See my commit description for details and also updated code solution

mfn:
@ulrichdahl can you please, for the sake of science and curiousity, check if https://github.com/rebing/graphql-laravel/pull/508 also works for your use-case?

ulrichdahl:
@mfn That was my own original solution (copied from an old post), but in the same way as I do not like including/requiring files inside classes or methods/functions. 
I see the original problem of declaring a class as the same as inclulding (should only be done in base namespace), and wanted to give a solution to that problem as well. However maybe the real solution would be to look at what code is using the Routing class and see if it could not use a common interface and make it agnostic to whether Lumen or Laravel is used??

ulrichdahl:
@mfn I believe I have found the best solution here. Remove the reference to Routing\Controller completely! It was not needed or used as far as I have been able to see and test. Please have a look and give feedback

mfn:
Ok, interesting fix 😅

My only concern is backwards compatibility and feature reliance. It's possible to:
- extend the controller
- switch your own via config
- rely on functionality provided by `\Illuminate\Routing\Controller`

If I remove it from the hierarchy, there's no way to get it back so I would potentially take away this functionality from anyone currently using it and provide no alternative 🤔

ulrichdahl:
@mfn Ya, I see that is a bit of a pickle.. I am not an expert in composer package building, so I do not know if there is an option to change stuff in basic includes based on which framework is used, thus being able to have a Lumen and Laravel controller/interface?
Would be a shame to begin forking this fine repo, based on this small issue.

mfn:
>  Would be a shame to begin forking this fine repo, based on this small issue.

First, no hard feelings, OSS & co 😄 

Counter question: would it be fine for you if I in fact merge https://github.com/rebing/graphql-laravel/pull/508 ? After all the discussion and new insights, I rather lean towards this solution. What do you think?

ulrichdahl:
@mfn none taken 😄 I am here to find the best solution, and I just had a thought based on your last comment... config!
Why not simply have 2 controllers?
There is already something specific to be done for Lumen installs, why not simply make a wrapper to be able to switch to a Lumen specific controller and then do the switch in the graphql config?
I can live with the solution in #508 so I leave it up to your capable hands to make a choice 😄

mfn:
Ok, please give me some time to think about this and I'll come back later to you, thanks!

mfn:
I'm closing this in favour of https://github.com/rebing/graphql-laravel/pull/508#issuecomment-677759110 which I just merged and left details as to why.

Very grateful for your research and patience here 🙇

PR #675, opened 2020-09-07
Css and JS updated to the latest GraphiQL version

albertcito:
## Summary
- Updated CSS and JS for GraphiQL section.
- Added a new title for GraphiQL section.

### DIFF
--- a/resources/views/graphiql.php
+++ b/resources/views/graphiql.php
@@ -12,7 +12,7 @@
     <meta name="robots" content="noindex" />
     <meta name="referrer" content="origin" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <title><?php echo $schema ? "$schema | " : ""; ?>GraphiQL</title>
+    <title><?php echo $schema ? "$schema | " : ''; ?>GraphiQL</title>
     <style>
       body {
         height: 100%;
--- a/resources/views/graphiql.php
+++ b/resources/views/graphiql.php
@@ -8,6 +8,11 @@
 <!DOCTYPE html>
 <html>
   <head>
+    <meta charset="utf-8" />
+    <meta name="robots" content="noindex" />
+    <meta name="referrer" content="origin" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <title><?php echo $schema ? "$schema | " : ""; ?>GraphiQL</title>
     <style>
       body {
         height: 100%;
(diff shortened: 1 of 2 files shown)


mfn (commented):
@albertcito thanks for the PR!

The title change looks absolutely legit / nice!

However, for the other changes: I'm not sure, can you point to the source of them?

When I adapted the files the last time I made a detailed README => https://github.com/rebing/graphql-laravel/blob/7918c746eed7d2c0bf35d66c7657f47a76cefead/resources/views/README.md

However re-reading it, it misses the most important information: where to exactly get the source from to sync it.

The best I then found is the initial commit https://github.com/rebing/graphql-laravel/commit/854c3a4d98c7fe7013a15d304ae13f08e4d8ee55 which details the (still valid!) URL where I got it from => https://github.com/graphql/graphiql/blob/v0.13.0/example/index.html

There is also a https://github.com/graphql/graphiql/blob/v0.13.1/example/index.html but when I checked it, there are changes actually.

After that, the file is gone as they've restructured the whole thing.

I think it's important to undertand where these changes come from, thanks!

albertcito:
This is the [current example] (https://github.com/graphql/graphiql/tree/main/examples/graphiql-cdn) to use GraphiQL using CDN. Since we are using the CDN urls, maybe we could use it and add this reference in the MD file for future updates. What do you think?

mfn:
Thanks for the link!

But the file there is quite different in nature if you look at the implementation.

So I'm not sure what the update now actually solves, as we're creating a Frankstein version out of the existing one from the `0.13.0` tag and now from this one, which diverged/evolved.

Are there any improvements with your PR _on the JS changes only_? Any new feature enabled due to this?

----
I rather lean towards having a "as reproducible as possible" graphiql file, which currently still is `0.13.0` plus some adjustments for the integration. Unless we come up with a completely different way …

My current TL;DR is:
- the `<title>` change seems fine and useful
- but the other changes don't as outlined above

What's you take how to move forward here?

albertcito:
What do you think about: 
- New title approved.
- Revert the JS changes. 

Is it ok?

mfn:
👍

resources/views/graphiql.php
```diff
@@ -150,4 +151,4 @@ function graphQLFetcher(graphQLParams) {
       );
     </script>
   </body>
-</html>
+</html>
```
mfn, line 154: I don't want to be a stickler, but here should be a newline 😅

mfn (approved):

mfn:
Thanks!

PR #680, opened 2020-09-30
Generate SyntaxError instead of Exception for empty queries

plivius:
## Summary
When providing a completely empty query, this triggers a hard (typecast) exception in `Rebing\GraphQL\GraphQLController` (either `$input` not being an array on L48 or `$query` being `null` on line L73). This PR facilitates a "graceful fallback" to a SyntaxError.

## Type of change

### Checklist
- [x] Amended CHANGELOG.md

### DIFF
--- a/src/GraphQLController.php
+++ b/src/GraphQLController.php
@@ -45,7 +45,10 @@ class GraphQLController extends Controller
 
         // Complete each query in order
         foreach ($inputs as $input) {
-            $completedQueries[] = $this->executeQuery($schema, $input);
+            $completedQueries[] = $this->executeQuery(
+                $schema,
+                ($input['query'] ?? null) ? $input : ['query' => '']
+            );
         }
 
         $data = $isBatch ? $completedQueries : $completedQueries[0];
--- a/src/GraphQLController.php
+++ b/src/GraphQLController.php
@@ -37,18 +37,25 @@ class GraphQLController extends Controller
             $schema = config('graphql.default_schema');
         }
 
-        // If a singular query was not found, it means the queries are in batch
-        $isBatch = ! $request->has('query');
-        $inputs = $isBatch ? $request->input() : [$request->input()];
+        // If a singular query parameter is not found, assume a batch request
+        $isBatch = false;
+        if ($request->input()) {
+            if ($request->has('query')) {
+                $inputs = [$request->input()];
+            } else {
+                $isBatch = true;
+                $inputs = $request->input();
+            }
+        } else {
+            // The request input is completely empty, assemble inputs accordingly
+            $inputs = [[]];
+        }
 
         $completedQueries = [];
 
         // Complete each query in order
         foreach ($inputs as $input) {
-            $completedQueries[] = $this->executeQuery(
-                $schema,
-                ($input['query'] ?? null) ? $input : ['query' => '']
-            );
+            $completedQueries[] = $this->executeQuery($schema, $input);
         }
 
         $data = $isBatch ? $completedQueries : $completedQueries[0];
@@ -61,7 +68,7 @@ class GraphQLController extends Controller
 
     protected function executeQuery(string $schema, array $input): array
     {
-        $query = $input['query'];
+        $query = $input['query'] ?? '';
 
         $paramsKey = config('graphql.params_key', 'variables');
         $params = $input[$paramsKey] ?? null;
--- a/tests/Database/EmptyQueryTest.php
+++ b/tests/Database/EmptyQueryTest.php
@@ -12,12 +12,12 @@ class EmptyQueryTest extends TestCaseDatabase
     use SqlAssertionTrait;
 
     /**
-     * @param $query
+     * @param string $query
      * @testWith    [""]
      *              [" "]
      *              ["#"]
      */
-    public function testEmptyQuery($query): void
+    public function testEmptyQuery(string $query): void
     {
         $this->sqlCounterReset();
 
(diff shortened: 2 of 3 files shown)


tests/Database/EmptyQueryTest.php
```diff
@@ -0,0 +1,30 @@
...
+use Rebing\GraphQL\Tests\TestCaseDatabase;
+
+class EmptyQueryTest extends TestCaseDatabase
+{
+    use SqlAssertionTrait;
+
+    /**
+     * @param string $query
+     * @testWith    [""]
+     *              [" "]
```
mfn, line 17: Please use a `@dataProvider`, this more in line with other tests, thank you!

src/GraphQLController.php
```diff
@@ -45,7 +45,10 @@ public function query(Request $request, string $schema = null): JsonResponse
 
         // Complete each query in order
         foreach ($inputs as $input) {
-            $completedQueries[] = $this->executeQuery($schema, $input);
+            $completedQueries[] = $this->executeQuery(
+                $schema,
+                ($input['query'] ?? null) ? $input : ['query' => '']
+            );
```
mfn, line 51: In your tests there's still always a `'query'`, so even if this fixes it, the tests don't proof it. Can you please extend them?

tests/Database/EmptyQueryTest.php
```diff
@@ -0,0 +1,30 @@
...
+declare(strict_types=1);
+
+namespace Rebing\GraphQL\Tests\Database;
+
+use Rebing\GraphQL\Tests\Support\Traits\SqlAssertionTrait;
+use Rebing\GraphQL\Tests\TestCaseDatabase;
+
+class EmptyQueryTest extends TestCaseDatabase
+{
+    use SqlAssertionTrait;
```
mfn, line 12: Can be removed (and the `sqlCounterReset` call), as no test is really using this

src/GraphQLController.php
mfn, line 51: I find this code quite hard to read, do you think this one will work too?
```suggestion
            $input['query'] = $input['query'] ?? '';
            $completedQueries[] = $this->executeQuery($schema, $input);
```
In the future this might get promoted to `??=` once 7.4 ever reaches the baseline.

mfn (changes requested):
Nice find and thanks for your first contribution 🥇 

There are still some smaller things to iron out, plus:
- can you please also change the base to the `5.x` branch?
  We can release this code faster then, I'll take care of the merge to master
- can you please add a changelog entry, too? You can add a `### Fixed` section

Thanks!

src/GraphQLController.php
plivius: Indeed, was trying too hard not to use `??=` ! 
I now moved the check to executeQuery().

plivius: This is indeed a good point! When examining this in more detail, I found that a fallback for a completely empty request (which was assumed to be a batch request) was probably also a good idea. Let me know if this is along the lines you have in mind.

src/GraphQLController.php
```diff
@@ -37,9 +37,19 @@ public function query(Request $request, string $schema = null): JsonResponse
...
+            if ($request->has('query')) {
+                $inputs = [$request->input()];
+            } else {
+                $isBatch = true;
+                $inputs = $request->input();
+            }
+        } else {
+            // The request input is completely empty, assemble inputs accordingly
+            $inputs = [[]];
+        }
```
mfn, line 52: Somehow I get the feeling this is overdoing things here.

When I revert this block back to the original version, only one test fails => the _first_ one in `\Rebing\GraphQL\Tests\Database\EmptyQueryTest::dataForEmptyQuery`:
```php
            // single request which is completely empty
            [
                false, [],
            ],
```
A request which is completely empty is, due to the current logic, expected to be a batch request. Yet one without any actual batches.

So the expected result to me is an empty array and not a syntax error.

Can you explain your goal of the new changes here?

mfn (commented):
My impression is that we should stick to a single fix here => the one in `\Rebing\GraphQL\GraphQLController::executeQuery` and adjust the test.

What do you think?

src/GraphQLController.php
mfn: > A request which is completely empty is, due to the current logic, expected to be a batch request. Yet one without any actual batches.

What I can imagine is that, if we would have the ability to _disable_ batch requests, _then_ being extra strict here makes sense.

That's a different change/PR though but one I could also get on board, as batch requests could have unforeseen consequences and security implications.

plivius: Addressing this incrementally indeed makes sense: The initial intention here was after all to resolve the hard exception. I have prepared a new branch based on 5.x, with the reverted changes. Let me know if you'd like me to submit that as a new PR:
https://github.com/rebing/graphql-laravel/compare/5.x...plivius:feature/proper-syntax-error-for-empty-query-5x

In the near future, it might still be a good idea to look into the case of the empty request (indeed perhaps dependent on enable/disable batch requests). Though the spec doesn't appear to be really explicit on this, if empty single/batch queries are considered incorrect, it would make sense the same is true for completely empty requests.

plivius: Hi @mfn, I've created a separate PR (#685) based on `5.x`, let me know if you would like to see any changes.

mfn: Thank you, I've noticed it already but currently lack enough time.

mfn:
@plivius closing this, as https://github.com/rebing/graphql-laravel/pull/685 was merged (i.e. the version containing my one feedback point regarding functionality).

PR #706, opened 2021-01-20
Support class based schemas

jasonvarga:
## Summary

This PR adds support for defining a class based schema.

Rather than defining the entire schema directly in the config file, like this:

```php
'schemas' => [
    'default' => [
        'query' => [
            QueryOne::class,
            QueryTwo::class
        ]
    ]
]
```

You'll be able to define it by specifying a class name:

```php
'schemas' => [
  'default' => \App\GraphQL\MySchema::class
]
```

This class should implement the new `ConfigConvertible` contract which will convert to an array you'd have previously used in the config file.

```php
use Rebing\GraphQL\Support\Contracts\ConfigConvertible;

class MySchema implements ConfigConvertible
{
    public function toConfig(): array
    {
        return [
            'query' => [
                QueryOne::class,
                QueryTwo::class
            ]
        ];
    }
}
```

The reason this would be useful for us (Statamic, a Laravel CMS package) is because we intend to be able to have an entire schema that users can put into their config. We don't want them to have to manually define the config.

Also, we intend to allow third parties to inject their own query classes into our schema. The schema class above is a bit simplified. Ours might look more like this:

```php
class Schema implements ConfigConvertible
{
    private static $customQueries = [];

    public function toConfig(): array
    {
        return [
            'query' => array_merge([
                NativeQueryOne::class,
                NativeQueryTwo::class
            ], static::$customQueries);
        ];
    }

    public static function addQuery($class)
    {
        static::$customQueries[] = $class;
    }
}
```

I'm not quite sure how you feel about the naming of things here though. I used `ConfigConvertible` and `toConfig()` because I saw a `TypeConvertible` interface with `toType()`.

The only _technically_ breaking change here is that if you already used a string, you'd previously get a GraphQL error like `cannot query field "some_query"` and now you'd get a `BindingResolutionException`. But your app wouldn't have been working in the first place so I don't think it really matters.

## Type of change

### Checklist
- [x] Update the README.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ CHANGELOG
 [Next release](https://github.com/rebing/graphql-laravel/compare/6.1.0...master)
 --------------
 
+### Added
+- Support for class based schemas [\#706 / jasonvarga](https://github.com/rebing/graphql-laravel/pull/706)
+
 2020-11-30, 6.1.0
 -----------------
 Same as 6.1.0-rc1!
--- a/README.md
+++ b/README.md
@@ -180,6 +180,39 @@ in addition to the global middleware. For example:
 ],
 ```
 
+#### Schema classes
+
+You may alternatively define the configuration of a schema in a class that implements `ConfigConvertible`.
+
+In your config, you can reference the name of the class, rather than an array.
+
+```php
+'schemas' => [
+    'default' => DefaultSchema::class
+]
+```
+
+```php
+namespace App\GraphQL\Schemas;
+
+use Rebing\GraphQL\Support\Contracts\ConfigConvertible;
+
+class DefaultSchema implements ConfigConvertible
+{
+    public function toConfig(): array
+    {
+        return [
+            'query' => [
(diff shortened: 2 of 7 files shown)


jasonvarga:
Not sure why the static analysis is failing. I didn't touch routes.php. Looks like it fails on master too.
The code style wants to make whacky migration class names. Also unrelated to anything in this PR.

🤔 🤷

mfn:
Hey @jasonvarga , thanks for the PR!

I'll check it out as soon as I find time, might take a while though. I wouldn't worry yet about the analyzer, I'll check this out later too.

jasonvarga:
Thanks!

mfn (approved):
Very nice, LGTM!

Also nice job with https://github.com/statamic/cms/pull/2982 😱 

Will merge & release as soon as everything is green

mfn:
@jasonvarga release as https://github.com/rebing/graphql-laravel/releases/tag/6.3.0 !

jasonvarga:
Awesome, thanks!

Issue #709, opened 2021-02-04
Support simplePagination


lamtranb:
Hi guys, here is my environment:

### Versions:
- graphql-laravel: 6.1.0
- Laravel Version: 8.26.1
- PHP Version: 7.4

### Question:
How can I use PaginationType with simplePagination ? When I use, it throws error:

`Argument 1 passed to Rebing\\GraphQL\\Support\\PaginationType::Rebing\\GraphQL\\Support\\{closure}() must be an instance of Illuminate\\Pagination\\LengthAwarePaginator, instance of Illuminate\\Pagination\\Paginator given, called in /app/vendor/webonyx/graphql-php/src/Executor/ReferenceExecutor.php on line 624`

Is there a workaround? 
Thanks.

mfn:
Closing as https://github.com/rebing/graphql-laravel/pull/715#issuecomment-813608749 was merged and released!

PR #715, opened 2021-02-18
support simple pagination

lamtranb:
## Summary
---
Hello, as #709 questioned, I think we should support simplePagination.
As Laravel doc mentioned here: https://laravel.com/docs/8.x/pagination#simple-pagination
> However, if you do not plan to show the total number of pages in your application's UI then the record count query is unnecessary.

If we have a large table then this will be a problem.

Type of change:

Checklist:
- [x] Existing tests have been adapted and new tests have been added
- [x] Update the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1564,7 +1564,7 @@ class PostsQuery extends Query
 {
     public function type(): Type
     {
-        return GraphQL::simplePaginate('posts');
+        return Type::nonNull(GraphQL::simplePaginate('posts'));
     }
 
     // ...
--- a/README.md
+++ b/README.md
@@ -1549,7 +1549,7 @@ Query `posts(limit:10,page:1){data{id},total,per_page}` might return
 Note that you need to add in the extra 'data' object when you request paginated resources as the returned data gives you
 the paginated resources in a data object at the same level as the returned pagination metadata.
 
-Simple Pagination will be used, if a query or mutation returns a `SimplePaginationType`.
+[Simple Pagination](https://laravel.com/docs/pagination#simple-pagination) will be used, if a query or mutation returns a `SimplePaginationType`.
 
 ```php
 namespace App\GraphQL\Queries;
--- a/README.md
+++ b/README.md
@@ -1549,6 +1549,37 @@ Query `posts(limit:10,page:1){data{id},total,per_page}` might return
 Note that you need to add in the extra 'data' object when you request paginated resources as the returned data gives you
 the paginated resources in a data object at the same level as the returned pagination metadata.
 
+Simple Pagination will be used, if a query or mutation returns a `SimplePaginationType`.
+
+```php
+namespace App\GraphQL\Queries;
+
+use Closure;
+use GraphQL\Type\Definition\ResolveInfo;
+use GraphQL\Type\Definition\Type;
+use Rebing\GraphQL\Support\Facades\GraphQL;
+use Rebing\GraphQL\Support\Query;
+
+class PostsQuery extends Query
+{
+    public function type(): Type
+    {
+        return GraphQL::simplePaginate('posts');
+    }
+
+    // ...
+
+    public function resolve($root, $args, $context, ResolveInfo $info, Closure $getSelectFields)
+    {
+        $fields = $getSelectFields();
+
+        return Post::with($fields->getRelations())
+            ->select($fields->getSelect())
+            ->simplePaginate($args['limit'], ['*'], 'page', $args['page']);
+    }
+}
+```
+
 ### Batching
 
 You can send multiple queries (or mutations) at once by grouping them together. Therefore, instead of creating two HTTP requests:
--- a/src/Support/SimplePaginationType.php
+++ b/src/Support/SimplePaginationType.php
@@ -1,0 +1,89 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Rebing\GraphQL\Support;
+
+use GraphQL\Type\Definition\ObjectType;
+use GraphQL\Type\Definition\Type as GraphQLType;
+use Illuminate\Pagination\Paginator;
+use Illuminate\Support\Collection;
+use Rebing\GraphQL\Support\Facades\GraphQL;
+
+class SimplePaginationType extends ObjectType
+{
+    public function __construct(string $typeName, string $customName = null)
+    {
+        $name = $customName ?: $typeName.'SimplePagination';
+
+        $config = [
+            'name'   => $name,
+            'fields' => $this->getPaginationFields($typeName),
+        ];
+
+        $underlyingType = GraphQL::type($typeName);
+        if (isset($underlyingType->config['model'])) {
+            $config['model'] = $underlyingType->config['model'];
+        }
+
+        parent::__construct($config);
+    }
+
+    /**
+     * @param string $typeName
+     *
+     * @return array<string, array<string,mixed>>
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -1549,6 +1549,37 @@ Query `posts(limit:10,page:1){data{id},total,per_page}` might return
...
+use GraphQL\Type\Definition\ResolveInfo;
+use GraphQL\Type\Definition\Type;
+use Rebing\GraphQL\Support\Facades\GraphQL;
+use Rebing\GraphQL\Support\Query;
+
+class PostsQuery extends Query
+{
+    public function type(): Type
+    {
+        return GraphQL::simplePaginate('posts');
```
mfn, line 1567: I think this can't be `null`? (Same for the existing example I would say, but let's stay focused on this PR)
```suggestion
        return Type::nonNull(GraphQL::simplePaginate('posts'));
```

README.md
```diff
@@ -1549,6 +1549,37 @@ Query `posts(limit:10,page:1){data{id},total,per_page}` might return
 Note that you need to add in the extra 'data' object when you request paginated resources as the returned data gives you
 the paginated resources in a data object at the same level as the returned pagination metadata.
 
+Simple Pagination will be used, if a query or mutation returns a `SimplePaginationType`.
```
mfn, line 1552: I think it would be nice to link to the Laravel docs here
```suggestion
[Simple Pagination](https://laravel.com/docs/pagination#simple-pagination) will be used, if a query or mutation returns a `SimplePaginationType`.
```

src/Support/SimplePaginationType.php
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * @param string $typeName
+     *
+     * @return array<string, array<string,mixed>>
+     */
+    protected function getPaginationFields(string $typeName): array
+    {
+        return [
+            'data'           => [
+                'type'        => GraphQLType::listOf(GraphQL::type($typeName)),
```
mfn, line 41: I don't think this field can ever return `null`:
```suggestion
                'type'        => Type::nonNull(GraphQLType::listOf(GraphQL::type($typeName))),
```

src/Support/SimplePaginationType.php
```diff
@@ -0,0 +1,89 @@
...
+                'description' => 'Current page of the cursor',
+                'resolve'     => function (Paginator $data): int {
+                    return $data->currentPage();
+                },
+                'selectable'  => false,
+            ],
+            'from'           => [
+                'type'        => GraphQLType::int(),
+                'description' => 'Number of the first item returned',
+                'resolve'     => function (Paginator $data): int {
```
mfn, line 66: Nullable field:
```suggestion
                'resolve'     => function (Paginator $data): ?int {
```

src/Support/SimplePaginationType.php
```diff
@@ -0,0 +1,89 @@
...
+                    return $data->firstItem();
+                },
+                'selectable'  => false,
+            ],
+            'to'             => [
+                'type'        => GraphQLType::int(),
+                'description' => 'Number of the last item returned',
+                'resolve'     => function (Paginator $data): int {
```
mfn, line 74: Nullable field:
```suggestion
                'resolve'     => function (Paginator $data): ?int {
```

mfn (changes requested):
Very nice @lamtranb !

Approach LGTM, can you check my feedback and also add a changelog entry?

PS: should phpstsan still complain, it's fine -> I can take care of that afterwards

Thanks!

src/Support/SimplePaginationType.php
lamtranb: Type::nonNull seems not work with ListOf. ListOf can return an empty list. I added a new test case for this. Or you mean this?
`GraphQLType::listOf(Type::nonNull(GraphQL::type($typeName)))`

src/Support/SimplePaginationType.php
lamtranb: I got this error from [https://github.com/rebing/graphql-laravel/runs/2265106812](https://github.com/rebing/graphql-laravel/runs/2265106812). Same with "to" field.
`Anonymous function never returns null so it can be removed from the  return typehint.   `

src/Support/SimplePaginationType.php
mfn: Good question!

> Type::nonNull seems not work with ListOf

Oh, but it should!

In this case I meant that the value returned by `resolve` cannot be `null`, ever, because even the closure has the typehint `Collection`.

This means there will always be a collection returned, never null, so wrapping the existing `GraphQLType::listOf(GraphQL::type($typeName)),` into non-null like `GraphQLType::nonNull(GraphQLType::listOf(GraphQL::type($typeName))),` should be possible!

(sorry I see in my feedback I just write `Type::nonNull` which is usually the alias used, but in this class it's `GraphQLType`, so maybe that's why it may look confusing).

> ListOf can return an empty list

Exactly! But "empty list" !== "null"!

> Or you mean this? `GraphQLType::listOf(Type::nonNull(GraphQL::type($typeName)))`

I would love to (always good to avoid null values, if possible), but I think a collection _containing_ null values is technically allowed (however unlikely), so therefore sadly no, didn't meant this one.

mfn (changes requested):
From https://github.com/rebing/graphql-laravel/pull/715#discussion_r606827182

>  Anonymous function never returns null so it can be removed from the return typehint. 

I guess the framework code _itself_ is contradicting, see https://github.com/laravel/framework/blob/6319d94ccaf8169b103813a82139814573ddef09/src/Illuminate/Pagination/AbstractPaginator.php#L321-L325

```php
    /**
     * Get the number of the first item in the slice.
     *
     * @return int
     */
    public function firstItem()
    {
        return count($this->items) > 0 ? ($this->currentPage - 1) * $this->perPage + 1 : null;
    }

```
The phpdoc (which phpstan obliges) mentions only `int` but in the code we clearly see it's possible to return `null` -> i.e. think about empty collections.

Same for `to`, in fact.

🤷‍♀️

Anyway, I "fixed" this one by adding it to phpstans baseline aka silencing it.

Can you look at the remaining feedback?

Thanks!

src/Support/SimplePaginationType.php
lamtranb: Hi @mfn 
Thanks for taking your time to review my code.

I updated the code into this:
`GraphQLType::nonNull(GraphQLType::listOf(GraphQL::type($typeName)))`
The tests failed.
I checked in the file tests/Database/SelectFields/PrimaryKeyTests/PrimaryKeySimplePaginationQuery.php
`
/** @var SelectFields $selectFields */
        $selectFields = $getSelectFields();
$selectFields->getSelect()
`
only return `posts.id`. If I remove GraphQLType::nonNull, it will return` posts.id, posts.title`. I think there is a bug in src/Support/SelectFields.php or miss check somewhere. Can you help me? I'm not familiar with this. Thanks very much.

mfn: TBH I've no idea.

I don't consider it a showstopper and I also didn't write the existing `PaginationType` which works _exactly_ the same way, so I actually do feel bad for side-tracking this 😞

mfn (approved):
LGTM, thank you very much!

mfn:
Released as https://github.com/rebing/graphql-laravel/releases/tag/7.0.1 !

PR #188, opened 2019-02-02
Add Lumen support

pierlon:
Now you can finally reach feature parity with [Folklore's](https://github.com/Folkloreatelier/laravel-graphql) package :smile:.

PR #722, opened 2021-03-12
[BREAKING CHANGE 🚨] Remove support for Lumen

mfn:
## Summary
Supporting Lumen without a dedicated framework for testing is a nightmare, any kind of fixes or changes are hard to verify. Also the readme never went beyond calling it "experimental".

So far the issues could be resolved, but due to it's niche application there are not many active contributions going on. And before Lumen supports jump on board, this project could rather use regular maintainers for the library and it's Laravel integration itself first and foremost.

The goal is to free this library from these requirements and allow it's further evolution without having to think about Lumen. 

This PR targets the next major release due to this breaking change. I encourage everyone requiring Lumen support of this library to create a fork.

---

Type of change:

Checklist:
- [x] Update the README.md

### Links
- Fixes https://github.com/rebing/graphql-laravel/issues/568


mfn:
@pierlon you made https://github.com/rebing/graphql-laravel/pull/188 , are you still using Lumen + graphql-laravel ?

Supporting Lumen over the last two years was the second most source of issues people reported and since there will never be a properly testing framework for it, I'm aiming for removing it.

pierlon:
@mfn No I no longer use it. It's unfortunate that it has caused so many issues but do feel free to remove it :slightly_smiling_face:.

mfn:
Pinging past contributors with Lumen specific code for awareness and feedback: @ulrichdahl @sglitowitzsoci @jstoks

Two ppl not pinged here already stated they're not using Lumen anymore (including the original author, see https://github.com/rebing/graphql-laravel/pull/722#issuecomment-817191348 )

ulrichdahl:
@mfn I use Lumen and this in many projects. If you will no longer maintain it for Lumen I will fork it and maintain my own fork.

mfn:
Thanks @ulrichdahl for your feedback!

Although I did anticipate ppl will just fork it, seeing it might indeed happen makes me feel a bit different about it.

I'll think a bit more about it and will not include in the next major version.

ulrichdahl:
@mfn I have actually changed my position since last 😄 so I am actually busy migrating all my projects to Laravel. Turns out that it is rather simple, only few hours of work is needed per project. Thank god it is simple projects.
Only 1 of them may stay in Lumen, but it is only a maybe. So could turn out to be ok for me too, right now I do not think I will fork anyway. 

But it looks like @Artem-Schander has forked and is changing it to be more Lumen ready out of the box.
https://github.com/EdwinDayot/graphql-laravel/commit/c3ece72d8bc2d02852692a3486809fc9a97f2575

Artem-Schander:
I'm alright. The few Lumen projects will be migrated to Laravel soon. But thanks for considering.

mfn:
> I have actually changed my position since last 😄

oh my, I'm glad I asked 😅

I'm well aware the people I pinged here don't reflect the real user base, but since we don't have many regular contributors, I was only concern with anyone who invested _something_ in the project.

I'm putting the PR back on track and merging it now.

---
For anyone arriving later here affected by this:
- please consider moving to Laravel
  - try to empathize the the work you need to invest for the switch (one time) vs. this project needs (long term)
  - TBH also seeing how seemingly easy ppl are able/willing/can switch to Lumen just relaxes by doubts about this move
- I'm not rejecting the idea to add support for it back in the future
  - But as I stated multiple times, the debt for it is unusually higher because there's no built-in way (like orchestra) to test things.
  - Right now my goal is to have a codebase I can understand and maintain, and the Lumen stuff doesn't help

I personally also think that the need for something like Lumen with the state of PHP 2021 isn't really there. It's not a coincidence there's a decline of contributions to https://github.com/laravel/lumen vs. the hype about different architectures like https://github.com/laravel/octane/ . I would rather make sure this library plays well with the latter than the former.

Thanks everyone for your feedback! 🙏

PR #724, opened 2021-03-16
Fix class based schemas within routes

jasonvarga:
## Summary

In #706 I added support for you to be able to generate a schema config by supplying a class name.
That worked when querying, but I've just noticed it doesn't work when generating the routes. 

i.e. `middleware` and `method` wouldn't be respected.

In `routes.php`, the routes are created by simply reading the config and looping over the schemas. If a class name is used, it needed to be converted to the array.

---

Type of change:

Checklist:
- [x] Update the README.md

### DIFF
--- a/src/GraphQL.php
+++ b/src/GraphQL.php
@@ -546,7 +546,7 @@ class GraphQL
         }
 
         /** @var ConfigConvertible $instance */
-        $instance = new $schema();
+        $instance = app()->make($schema);
 
         return $instance->toConfig();
     }
--- a/src/GraphQL.php
+++ b/src/GraphQL.php
@@ -514,7 +514,7 @@ class GraphQL
     }
 
     /**
-     * @return array<array>
+     * @return array<array|Schema>
      */
     public static function getNormalizedSchemasConfiguration(): array
     {
@@ -528,13 +528,17 @@ class GraphQL
     }
 
     /**
-     * @param  mixed  $schema
-     * @return array<array>
+     * @param  Schema|array<array>|string|null  $schema
+     * @return Schema|array<array>
      */
-    protected static function getNormalizedSchemaConfiguration($schema): array
+    protected static function getNormalizedSchemaConfiguration($schema)
     {
-        if (! is_string($schema)) {
+        if (is_array($schema) || $schema instanceof Schema) {
             return $schema;
+        }
+
+        if (is_null($schema)) {
+            return [];
         }
 
         if (! class_exists($schema)) {
--- a/src/GraphQL.php
+++ b/src/GraphQL.php
@@ -500,7 +500,7 @@ class GraphQL
      * @param  array|string|null  $schema
      * @return array|Schema
      */
-    public function getSchemaConfiguration($schema)
+    protected function getSchemaConfiguration($schema)
     {
         $schemaName = is_string($schema) ? $schema : config('graphql.default_schema', 'default');
 
--- a/src/GraphQL.php
+++ b/src/GraphQL.php
@@ -531,7 +531,7 @@ class GraphQL
      * @param  Schema|array<array>|string|null  $schema
      * @return Schema|array<array>
      */
-    protected static function getNormalizedSchemaConfiguration($schema)
+    public static function getNormalizedSchemaConfiguration($schema)
     {
         if (is_array($schema) || $schema instanceof Schema) {
             return $schema;
--- a/src/GraphQL.php
+++ b/src/GraphQL.php
@@ -513,7 +513,10 @@ class GraphQL
         return static::getNormalizedSchemaConfiguration($schema);
     }
 
-    public static function getNormalizedSchemasConfiguration()
+    /**
+     * @return array<array>
+     */
+    public static function getNormalizedSchemasConfiguration(): array
     {
         return array_filter(array_map(function ($schema) {
             try {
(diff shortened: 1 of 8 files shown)


mfn:
We cannot use the Facade in the routes: this would fully have and unconditionally boot GraphQL which is not acceptable for the routes.

Seems the solution needs a stateless variant of `getSchemaConfiguration` or something.

To goal would be either to:
- avoid booting GraphQL at all
  or
- only take the hit in the case such "class based schemas" are used (which should be added to the README too).

At least, these are my first quick thoughts on this.

jasonvarga:
No worries. I made a draft so I can run the test suite. I'll fix it up.

src/GraphQL.php
```diff
@@ -510,12 +510,39 @@ protected function getSchemaConfiguration($schema)
...
             return $schema;
         }
 
+        if (! class_exists($schema)) {
+            throw new SchemaNotFound('Schema class '.$schema.' not found.');
+        }
+
         /** @var ConfigConvertible $instance */
-        $instance = app()->make($schema);
+        $instance = new $schema();
```
mfn, line 545: We should not change that, btw

jasonvarga: Agreed it would be nice to keep that. I don't know how necessary it is to have dependencies injected for this thing.
But I know removing it would make it a breaking change. 🤔 Bummer. It's only been there for a moment and I'm probably the only person that's tried to use the feature. 😆 

I changed this to do a regular `new`ing up of the class so that the app doesn't have to get booted.

mfn: TBH I'm not so much about the breaking change, hardly imagine people jumped on this new feature released a few days ago 😅

No, to me it's really more about the idiomaticity (sp?), what to expect in a Laravel based library.

jasonvarga: I took your comment about not booting a little too far. You meant just avoid booting GraphQL. Resolving a class through the container wouldn't trigger booting GraphQL. The app is already booting anyway so using the container there is fine.
Silly me. 😬

jasonvarga:
This is ready for review whenever you are. (In case you didn't get the notification - not hassling you)

jasonvarga:
I re-requested a review, just to see if you get a notification. 🤷 yes GitHub things.

Thanks!

src/GraphQL.php
```diff
@@ -510,10 +510,41 @@ protected function getSchemaConfiguration($schema)
...
+                return null;
+            }
+        }, config('graphql.schemas')));
+    }
+
+    /**
+     * @param  Schema|array<array>|string|null  $schema
+     * @return Schema|array<array>
+     */
+    protected static function getNormalizedSchemaConfiguration($schema)
```
mfn, line 534: I suggest to stay consistent with the visibility as this goes together with `getNormalizedSchemasConfiguration`:
```suggestion
    public static function getNormalizedSchemaConfiguration($schema)
```

mfn (commented):
Hmm, so I tried to run this in a private project and basically GraphQL totally broke; got about 5k tests and none of them work anymore 😏 

Some issue with type resolution:
```
array (
  'errors' => 
  array (
    0 => 
    array (
      'status' => '500',
      'code' => '0',
      'title' => 'Rebing\\GraphQL\\Exception\\TypeNotFound: Type <SomeType> not found.
Check that the config array key for the type matches the name attribute in the type\'s class.
It is required when \'lazyload_types\' is enabled',
      'meta' => 
      array (
        'stacktrace' => '#0 /project/vendor/rebing/graphql-laravel/src/GraphQL.php(193): Rebing\\GraphQL\\GraphQL->getType(\'SomeType\', false)
#1 /project/vendor/laravel/framework/src/Illuminate/Support/Facades/Facade.php(261): Rebing\\GraphQL\\GraphQL->type(\'SomeType\')
#2 /project/app/GraphQL/Queries/SomeQuery.php(34): Illuminate\\Support\\Facades\\Facade::__callStatic(\'type\', Array)
```
I haven't yet dug into it.

jasonvarga:
Well that's not great. 🤔 I'll try to work that out.

mfn:
@jasonvarga it was a false alarm I think 😬

This PR wasn't up2date with master and when I tried it in a project, it was already relying on other improvements in master which are not yet in here and thus broke I think.

I therefore resolved the conflict and merged master, will try again now.

mfn (approved):
Thanks, LGTM!

Especially that it does not have any additional penalty unless you use the class based schema looks well done to me, thank you!

mfn:
Released as https://github.com/rebing/graphql-laravel/releases/tag/6.5.0 !

jasonvarga:
Thank you as well!

src/GraphQL.php
```diff
@@ -503,10 +503,41 @@ protected function getSchemaConfiguration($schema)
...
+
+    /**
+     * @return array<string, array|Schema>
+     */
+    public static function getNormalizedSchemasConfiguration(): array
+    {
+        return array_filter(array_map(function ($schema) {
+            try {
+                return static::getNormalizedSchemaConfiguration($schema);
+            } catch (SchemaNotFound $e) {
```
mfn, line 517: @jasonvarga do you remember why you added this here?

I was playing around with some refactoring in other parts when I saw this and I wondered: if a schema is broken, aka not found, why would I silence the schema here?

Aka: a broken schema throws an exception for a purpose, it should crash the app?

jasonvarga: Good question. 😬 

In 6.2.0, if you had an invalid class (or just a string, really, since class names weren't supported yet) it would not throw an exception. I think I just made it continue to _not_ throw an exception to keep the behavior consistent and not a breaking change.

mfn: Thanks!

I decided to change it (https://github.com/rebing/graphql-laravel/pull/766) because somehow it didn't make sense to have an invalid configuration being silently skipped (as this is used for generated the routes, you're suddenly left in the dark why a certain schema simply "isn't there").

jasonvarga: No arguments here!

PR #738, opened 2021-04-09
Allow disabling batched requests

mfn:
## Summary

- See https://www.apollographql.com/blog/batching-client-graphql-queries-a685f5bcd41b/ for pro/con
- Stays enabled by default but can be disabled if desired
- will return a GraphQL batch compatible error response
  i.e. each batch will return the same error, however not more than 100 such errors are returned

---

Type of change:

Checklist:
- [x] Update the README.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ CHANGELOG
 [Next release](https://github.com/rebing/graphql-laravel/compare/7.1.0...master)
 --------------
 
+### Added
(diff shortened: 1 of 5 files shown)


mfn:
Github Actions has some issues, the state does not reflect the current runs. Can be seen via https://github.com/rebing/graphql-laravel/actions/runs/734334562 where everything is green

mfn:
Note: I've decided to change the config slightly and it's now `batching.enable` instead of `batching` (see https://github.com/rebing/graphql-laravel/commit/89ba2f40880963724e93ed54ccc8fc5f8a0d89b5 )

PR #740, opened 2021-04-11
apq: parse query ahead of caching

mfn:
## Summary
Once a query is retrieved from the cache, it's not necessary to parse it
again.

This builds upon https://github.com/rebing/graphql-laravel/pull/701

The reason for keeping the original source query is to still allow calling `queryContext()` with it (otherwise there would indeed not be much reason to keep it).

### Notes
- ~Lumen integration tests are broken, but this is unrelated and already happens since https://github.com/rebing/graphql-laravel/commit/53248e2cfa5c3de78562f7f5b5513c13933c8dc0~ fixed with https://github.com/rebing/graphql-laravel/commit/85a7e885107af47edff2cf753d1e1bd0457df30c

---

Type of change:

Checklist:
- [x] Update the README.md

### DIFF
--- a/src/GraphQLController.php
+++ b/src/GraphQLController.php
@@ -105,7 +105,7 @@ class GraphQLController extends Controller
                 'query' => $query,
                 'parsedQuery' => $parsedQuery,
             ] = $this->handleAutomaticPersistQueries($schemaName, $params);
-        } catch (AutomaticPersistedQueriesError $e) {
+        } catch (AutomaticPersistedQueriesError | Error $e) {
             return $graphql
                 ->decorateExecutionResult(new ExecutionResult(null, [$e]))
                 ->toArray($debug);
--- a/src/GraphQLController.php
+++ b/src/GraphQLController.php
@@ -7,6 +7,7 @@ use Exception;
 use GraphQL\Error\DebugFlag;
 use GraphQL\Error\Error;
 use GraphQL\Executor\ExecutionResult;
+use GraphQL\Language\Parser;
 use GraphQL\Server\Helper;
 use GraphQL\Server\OperationParams;
 use GraphQL\Server\RequestError;
@@ -98,7 +99,12 @@ class GraphQLController extends Controller
         }
 
         try {
-            $query = $this->handleAutomaticPersistQueries($schemaName, $params);
+            // In case of an APQ cache hit, parsedQuery contains the AST of the provided query
+            // and is subsequently used to speed up the execution.
+            [
+                'query' => $query,
+                'parsedQuery' => $parsedQuery,
+            ] = $this->handleAutomaticPersistQueries($schemaName, $params);
         } catch (AutomaticPersistedQueriesError $e) {
             return $graphql
                 ->decorateExecutionResult(new ExecutionResult(null, [$e]))
@@ -106,7 +112,7 @@ class GraphQLController extends Controller
         }
 
         return $graphql->query(
-            $query,
+            $parsedQuery ?? $query,
             $params->variables,
             [
                 'context' => $this->queryContext($query, $params->variables, $schemaName),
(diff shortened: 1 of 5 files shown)


mfn:
@illambo what do you think about this one?

src/GraphQLController.php
```diff
@@ -96,15 +97,20 @@ static function (RequestError $err): Error {
         }
 
         try {
-            $query = $this->handleAutomaticPersistQueries($schemaName, $params);
+            // In case of an APQ cache hit, parsedQuery contains the AST of the provided query
+            // and is subsequently used to speed up the execution.
```
illambo, line 101: I feel old :smiley:

illambo (commented):
super! I'll play it tomorrow

crissi:
Does this persisted query implementation also optimize normal queries sent?

I send a normal graphql query and then send one again is the parsing skipped then for the second? Or do we only get the benefits when we send a hashed query?

illambo:
Hello I'm starting to play with it a bit.

About:
> The reason for keeping the original source query is to still allow calling queryContext() with it (otherwise there would indeed not be much reason to keep it).

Testing with:
https://github.com/rebing/graphql-laravel/blob/f8e5db70696d717b275bcc98f7cd1c7528509e1d/tests/Unit/AutomatedPersistedQueriesTest.php#L93
I have found that the query seems be accessed via `->definitions->offsetGet(0)->loc->source->body` in the parsed query (`$datum['parsedQuery']`).

Maybe something can be done to remove the query from the cache and clean up the method.

illambo:
Hi @crissi you need to use the apq [protocol](https://github.com/apollographql/apollo-link-persisted-queries#protocol) to get the benefit.
For example apollo client give you this behavior transparent for u, see [createPersistedQueryLink](https://www.apollographql.com/docs/apollo-server/performance/apq/#apollo-client-setup).

src/GraphQLController.php
```diff
@@ -96,15 +97,20 @@ static function (RequestError $err): Error {
...
+            [
+                'query' => $query,
+                'parsedQuery' => $parsedQuery,
+            ] = $this->handleAutomaticPersistQueries($schemaName, $params);
         } catch (AutomaticPersistedQueriesError $e) {
```
illambo, line 106: Perhaps it is better to add error handling related to the method `Parser::parse($query)` (like GraphQL\Error\SyntaxError).
What do you think about it?

src/GraphQLController.php
```diff
@@ -126,11 +132,17 @@ protected function queryContext(string $query, ?array $variables, string $schema
...
      *       to the negotiation protocol.
+     * @return array{query:string,parsedQuery:?\GraphQL\Language\AST\DocumentNode}
      */
-    protected function handleAutomaticPersistQueries(string $schemaName, OperationParams $operation): string
+    protected function handleAutomaticPersistQueries(string $schemaName, OperationParams $operation): array
     {
         $query = $operation->query;
 
+        $datum = [
+            'query' => $query,
```
illambo, line 142: see [comment](https://github.com/rebing/graphql-laravel/pull/740#issuecomment-818117141), maybe it can clean something

illambo:
For now I can say: great! 
I think a beta period might be useful to collect some opinion / feedback.

src/GraphQLController.php
mfn: Good point, will do!

src/GraphQLController.php
mfn: Good hint, will investigate!

src/GraphQLController.php
mfn: This is handled now + added a test, again thanks!

src/GraphQLController.php
mfn: >  I have found that the query seems be accessed via ->definitions->offsetGet(0)->loc->source->body in the parsed query ($datum['parsedQuery']).

For now I decided against this approach: this requires lots of "internal detail knowledge" about the AST, which I'd rather not add as I don't see much overhead storing the original query (compared to the much bigger AST already added, too).

mfn:
>  I think a beta period might be useful to collect some opinion / feedback.

TBH I don't plan to: it's a new feature and has to be deliberately enabled -> so I don't see much risk involved here. If APQ is not enabled, the execution flow isn't touched.

mfn:
@illambo what do you think about the current state here?

src/GraphQLController.php
illambo: Ok, I was also quite undecided about it.

illambo:
Everything seems ok to me and I am very happy with the PR!
I think you can proceed with the integration if you have no other thoughts about it.

mfn:
>  I think you can proceed with the integration if you have no other thoughts about it.

Thank you!

Issue #465, opened 2019-08-26
Schema name can't contain hyphens


AlexanderSix:
### Versions:
- graphql-laravel Version: 2.0.1
- Laravel/Lumen Version: 5.8.33
- PHP Version: 7.2.21

### Description:

At work, we recently added this package to a project that we have started for a client (and we're loving the package, by the way), but we bumped into some problems when trying to create a second schema to read from. We were getting a `NotFoundHttpException` each time we tried to access the schema via the api.

Our first schema was called "read", but the one that was giving us problems was called "read-public".

After some trial-and-error, we found out that replacing the hyphen (-) with an underscore (_) fixed the problem. That said, is this a bug, or is this intended functionality? If it **is** how the package is supposed to work, maybe we should add something to the "Schema" section of the documentation that explicitly says this.

If you want to go the route of adding to the documentation, I would be happy to add that in a PR when I have a spare few minutes over the next few days.

### Steps To Reproduce:

1. Add a schema to the 'schemas' array in graphql.php that has a hyphen(-) in the name
2. Remove any middlewares that are attached to the query
3. Try to hit that endpoint with an API request
4. You should receive a `NotFoundHttpException` on the response

mfn:
Indeed, https://github.com/rebing/graphql-laravel/pull/467 shows it doesn't work.

No idea why and, yes, my expectation would be that it should work 🤷‍♀️

mfn:
Can you help debugging this?

mfn:
It seems the `->where()` calls sprinkled all over `src/routes.php` cause this. When I comment them out, the test from the other PR works (or fails, however you view it: the exception is not thrown anymore).

mfn:
Correct, in terms of regex, they do.

I've never used the `->where()` feature myself, literally no idea why it's used here.

Seems there's also no test for it, I can remove them all and the whole suite passes (except the new test specifically for this scenario: because it would fail because it "works").

(I also wonder how they work when they've no delimiter anywhere 🤔 but I see it gets passed to the symfony route class, so 🤷‍♀️)

This is the commit introducing the feature: https://github.com/rebing/graphql-laravel/commit/c44666b7746213851fc2290313618cf1c18f2a52

The commit message does not give much away, also no references to anything.

The added `{` and `}` turns the previous literal part into a route parameter. Then it seems the `->where()` is used to define matching rules for the parameter which is ... literally the same name, again.

Would appreciate help from someone more knowledgeable in this area 🤷‍♀️

PR #757, opened 2021-04-24
[BREAKING CHANGE 🚨] Rewrite and simplify routing

mfn:
## Summary
- rewrote the routing code from scratch: the group routing is kept but
  the setup for each schema is vastly simplified
- moved all routing related configuration into the `route` top level
  configuration key
- register only one route per schema (and not one per method)
  - plus one for the default schema on the route group prefix
  - same for graphiql
- remove support for different routes for schema query
  and schema mutations.
- to keep supporting multi-level schemas, e.g. schemas looking like
  `multi/level`, we don't capture the schema name from the group but
  resolve it in the controller from the request.
  The problem is that `/` are not possible to be caught that way.
  OTOH it simplified the route parameter logic in `query()`
  (i.e. it was removed)

For the longest time, the whole routing code was a real eyesore. This library is a bit unique in that it supports multiple schemas which not all do, but I don't think we need to be able to have potential "route for queries in a schema" and a different route for "mutations in a schema"; each which could target different controller actions.

Fun fact: this Fixes https://github.com/rebing/graphql-laravel/issues/465 (although this was not intentional, it's a side effect of removing the complexity).

To handle such different behaviour one could:
- use multi-level routes
- use the resolver middleware

## The breaking changes
- config options have been shuffled around in an incompatible way
- different routes for `query` and `mutation` within a schema not explicitly supported anymore
- the internal route names have changed: if you're using them in your application, it's not `graphql.query` anymore but just `graphql` or `graphql.your_non_default_schema`

---

Type of change:

Checklist:
- [x] Update the README.md

### DIFF
--- a/src/GraphQL.php
+++ b/src/GraphQL.php
@@ -479,28 +479,6 @@ class GraphQL
     }
 
     /**
-     * Check if the schema expects a nest URI name and return the formatted version
-     * Eg. 'user/me'
-     * will open the query path /graphql/user/me.
-     */
-    public static function routeNameTransformer(string $name, string $schemaParameterPattern, string $queryRoute): string
-    {
-        $multiLevelPath = explode('/', $name);
-        $routeName = null;
-
-        if (count($multiLevelPath) > 1) {
-            foreach ($multiLevelPath as $multiName) {
-                $routeName = !$routeName ? null : $routeName . '/';
-                $routeName =
-                    $routeName
-                    . preg_replace($schemaParameterPattern, '{' . $multiName . '}', $queryRoute);
-            }
-        }
-
-        return $routeName ?: preg_replace($schemaParameterPattern, '{' . $name . '}', $queryRoute);
-    }
-
-    /**
      * @param array|string|null $schema
      * @return array|Schema
      */
--- a/src/GraphQLServiceProvider.php
+++ b/src/GraphQLServiceProvider.php
@@ -37,9 +37,7 @@ class GraphQLServiceProvider extends ServiceProvider
      */
     protected function bootRouter(): void
     {
-        if (config('graphql.routes')) {
-            include __DIR__ . '/routes.php';
-        }
+        $this->loadRoutesFrom(__DIR__ . '/routes.php');
     }
 
     /**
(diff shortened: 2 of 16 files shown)


src/GraphQL.php
```diff
@@ -478,28 +478,6 @@ public static function handleErrors(array $errors, callable $formatter): array
         return array_map($formatter, $errors);
     }
 
-    /**
-     * Check if the schema expects a nest URI name and return the formatted version
-     * Eg. 'user/me'
-     * will open the query path /graphql/user/me.
-     */
-    public static function routeNameTransformer(string $name, string $schemaParameterPattern, string $queryRoute): string
```
mfn, line 486: Not necessary anymore => removed

src/GraphQLServiceProvider.php
```diff
@@ -37,9 +37,7 @@ public function boot(): void
      */
     protected function bootRouter(): void
     {
-        if (config('graphql.routes')) {
-            include __DIR__ . '/routes.php';
-        }
+        $this->loadRoutesFrom(__DIR__ . '/routes.php');
```
mfn, line 40: This change leads to honoring the route cache

tests/Database/AuthorizeArgsTests/AuthorizeArgsTest.php
```diff
@@ -11,8 +11,6 @@ protected function getEnvironmentSetUp($app): void
     {
         parent::getEnvironmentSetUp($app);
 
-        $app['config']->set('graphql.controllers', GraphQLController::class . '@query');
-
```
mfn, line 15: The config has been renamed, but this actual setting had no impact anyway => removed

tests/Unit/ConfigTest.php
```diff
@@ -21,13 +21,6 @@ class ConfigTest extends TestCase
     protected function getEnvironmentSetUp($app): void
     {
         $app['config']->set('graphql', [
-            'prefix' => 'graphql_test',
-
-            'routes' => [
-                'query' => 'query/{graphql_schema?}',
-                'mutation' => 'mutation/{graphql_schema?}',
-            ],
```
mfn, line 29: This feature is not supported anymore.

The `prefix` is already removed, because it was otherwise not used in any of the remaining tests.

tests/Unit/SchemaHyphenInPathTest.php
```diff
@@ -34,8 +33,6 @@ public function testWithHyphen(): void
             'query' => $graphql,
         ]);
 
-        $this->expectException(NotFoundHttpException::class);
-
```
mfn, line 38: Accidental bugfix 😄

Issue #753, opened 2021-04-20
Pre hook/middleware


crissi:
Gives us the ability to change/check the query received before it is being parsed.

Cases:
 * detect unused variables ( mfn already implented this )
 * change a query type, ex. from int to string, without breaking the api for consumers
 * change variables to conform to the newest webonyx release (has more strict variable checking).

mfn:
> Gives us the ability to change/check the query received before it is being parsed.

Any proposal as to how to do it?

> detect unused variables ( mfn already implented this )

Maybe an architecture which can incorporate the manual code I added; would be glad for that.

> change variables to conform to the newest webonyx release (has more strict variable checking).

Do you have more details?

crissi:
I think it could be implemented a lot like the middleware already in existent, but since we dont have the query name ( i think) it would probably be in a global config.

_change variables to conform to the newest webonyx release (has more strict variable checking)._
Strict coercion of scalar types in https://github.com/webonyx/graphql-php/releases/tag/v14.0.0

mfn:
The only worrying thing to me right now is to introduce yet another middleware/hook layer 😅

> _change variables to conform to the newest webonyx release (has more strict variable checking)._
> Strict coercion of scalar types in https://github.com/webonyx/graphql-php/releases/tag/v14.0.0

Oh, I remotely remember. What are you suggest here? A hook which is in fact able to relax the coercion strictness for input?

mfn:
We would have three layers of middleware:
- http middleware (aka what we already have with Laravel)
- execution middleware (new feature?)
- resolver middleware (implemented)

All work similarly in concept but all have different signatures and serve different purposes.

I guess this "execution middleware" would work within _each_ result of a batch, distinctively ? https://github.com/rebing/graphql-laravel/blob/65270611e1c59c9303ecb31600a6267ddee68ff1/src/GraphQLController.php#L65-L70 and would be defined on the "schema level" (a bit like the http middleware but unlike the resolver middleware; though nothing would prevent from making the resolve middleware also on the schema level aka global possible etc.)

Basically around the notion of `executeQuery($schemaName, $operation)` we would have something in between

However we would need additional flexibility: I can imagine that the string GraphQL query might get parsed into its AST and then also worked with during that phase:
[...]

PR #762, opened 2021-05-03
Execution middleware

crissi:
## Summary
Execution hook/middleware based on https://github.com/rebing/graphql-laravel/issues/753

### TODO
 - [x] convert current middleware like functions to use this instead
 - [x] docs
 - [x] test for batched requests
 - [x] document breaking changes in more detail
 - [x] test per-schema execution middleware

---

Type of change:

Checklist:
- [x] Update the README.md


crissi:
@mfn can you take a look when you have some time!

src/GraphQL.php
```diff
@@ -118,6 +119,17 @@ public function schema($schema = null): Schema
      */
     public function query($query, ?array $variables = [], array $opts = []): array
     {
+        $middlewareResponse = app()->make(Pipeline::class)
+            ->send([$query, $variables])
```
mfn, line 123: I think we should add `$opts` here too.

Or, well at least the most important bit would be the `schemaName`.

TBH I despise the whole signature of this `query` method, it just does not make any sense to me.

I think the one from [\Rebing\GraphQL\GraphQLController::executeQuery](https://github.com/rebing/graphql-laravel/blob/26302c195b5dbd29d5d79e72b61cc783da047a39/src/GraphQLController.php#L55) makes much more sense

with the addition of being able to carry over the _parsed query_ (either freshly parsed or retrieved e.g. from APQ).

src/GraphQL.php
```diff
@@ -151,6 +163,14 @@ public function queryAndReturnResult($query, ?array $variables = [], array $opts
         return GraphQLBase::executeQuery($schema, $query, $rootValue, $context, $variables, $operationName, $defaultFieldResolver);
     }
 
+    /**
+     * @return array<string>
+     */
+    public function executionMiddleware(): array
+    {
+        return config('graphql.execution_middleware', []);
```
mfn, line 171: I think it's important we consider the schema name too (*), something like:
```php
return
    config("graphql.$schemaName.execution_middleware") ??
    config('graphql.execution_middleware') ??
    [];
```

(*) which isn't in this context yet, I know

src/Support/ExecutionMiddleware.php
```diff
@@ -0,0 +1,32 @@
...
+     * @param array<string,mixed> $arguments
+     * @return Closure|array<mixed>
+     */
+    public function resolve(array $arguments, Closure $next)
+    {
+        return $this->handle(...$arguments, ...[
+            function (...$arguments) use ($next) {
+                return $next($arguments);
+            },
+        ]);
```
mfn, line 30: I don't think I'll ever understand this brainf..kery of code 😛

mfn (commented):
# I like 👍 

Not sure were we end up with this, but I see this as a promising start! Also: so little code 😱 

I think we should try hard with the actual data we pass and I fully accept further breaking changes for the upcoming 8 release cycle; it's the best time now.

Feel free to react to the feedback already, but I definitely need more time to play around with this. Now that we've something working, it's really best probably to try / see implement:
- APC
- detect_unused_variables
- …

to better understand what is / what is not possible and what's missing etc.

Thanks for the initiative 🙏

mfn:
Btw, why are tests not running? I know there were some recent GHA changes, but you're not a first time contributor and I also don't see a button for approval of the requests:
image

Or are tests on public repos in _draft_ mode not executed anymore? 🤔

src/Support/ExecutionMiddleware.php
mfn: Ah, it "clicked" now; essentially this is this:
```php
        $arguments[] = function (...$arguments) use ($next) {
            return $next($arguments);
        };

        return $this->handle(...$arguments);
```

But once unpacking is used, you can't have positional arguments anymore.

mfn:
Just thinking out loud, no need to do anything right now:

No matter how we name this middleware, we probably should rename `\Rebing\GraphQL\Support\Middleware` into `ResolverMiddleware`. I wonder what the guide of "where to place things" should be. It would not be unlikely a project will use a mix of resolver and execution middlewares; and then where should we but them? Should we stick to `\GraphQL\Middleware` https://github.com/rebing/graphql-laravel/blob/8b156d5b98dee7e3fb86c62f42253a1d647de288/src/Console/MiddlewareMakeCommand.php#L21 

for both "kinds" of middlewares?

Maybe moving both to `\Rebing\GraphQL\Middleware\` would be a good idea (`ResolverMiddleware` and `ExecutionMiddleware`)

src/GraphQL.php
```diff
@@ -118,6 +119,17 @@ public function schema($schema = null): Schema
...
     {
+        $middlewareResponse = app()->make(Pipeline::class)
+            ->send([$query, $variables])
+            ->through($this->executionMiddleware())
+            ->via('resolve')
+            ->thenReturn();
+
+        if (1 === count($middlewareResponse)) {
+            return $middlewareResponse;
+        }
```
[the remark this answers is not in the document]
crissi: this catches, if you returning directly in the middleware.

After thinking about it this might not be enough to know for sure that it was a response that has been returned, will investigate further

crissi:
> Feel free to react to the feedback already, but I definitely need more time to play around with this. Now that we've something working, it's really best probably to try / see implement:
> 
> * APC
> * detect_unused_variables
> * …
> 
> to better understand what is / what is not possible and what's missing etc.
> 

I will try to move them to the middleware

mfn:
> Btw, why are tests not running?

Aha! The tests are entirely running only on the fork: https://github.com/crissi/graphql-laravel/actions

Uh, I think I got an idea: maybe because I removed the `push` -> will try this now.

mfn:
It works 🥳 

Will fix the phpstan error

src/Support/ExecutionMiddleware.php
crissi: haha me neither, just copied: :smile:

crissi:
I tried to move the apc part but got into problems with the queryContext missing a query string, to fix this the query context needs to be added to the Graphql class instead of the controller. But maybe we can let that be at is.

mfn:
> I tried to move the apc part but got into problems with the queryContext missing a query string, to fix this the query context needs to be added to the Graphql class instead of the controller. But maybe we can let the apc be at is!?

I was working the better half of the evening now on this.

I broke almost half the tests.

Basically I was able to turn _everything_ into middlewares:
- `AddAuthUserContextValueMiddleware`
- `AutomaticPersistedQueriesMiddleware`
- `UnusedVariablesMiddleware`
- `ValidateOperationParamsMiddleware`

but the cake is taken by turning the _actual GraphQL execution_ into its own middleware: `GraphqlExecutionMiddleware` ❗ 
At first it sounds ridiculous to do this, but when I saw the latest version before my refactoring:
```php
        $middlewareResponse = $this->app->make(Pipeline::class)
            ->send([$query, $variables, $opts])
            ->through($this->executionMiddleware($schemaName))
            ->via('resolve')
            ->thenReturn();

        if ($middlewareResponse instanceof ExecutionResult) {
            return $middlewareResponse;
        }

        [$query, $variables] = $middlewareResponse;

        return GraphQLBase::executeQuery($schema, $query, $rootValue, $context, $variables, $operationName, $defaultFieldResolver);
```
seeing how query/variables are returned from the pipeline and passed on, setting how we've to special handle ExecutionResult, it dawned on me that all this special case handling goes away the moment the actual execution is just a middleware too.

One challenge was to accommodate for the fact that the GraphqL query has to be parsed and may also be changed due to the APQ requirements. I created `OperationParams` for this, extending the one from the graphql-php library but adding support for getting(+parsing)/setting the (parsed) query.

But the fact how much code was deleted from GraphQLController and GraphQL is astonishing; basically each method removed was turned into a middleware.

Unfortunately this required big changes to what parameter we pass around, therefore the signatures from
- `\Rebing\GraphQL\GraphQL::query`
  and
- `\Rebing\GraphQL\GraphQL::queryAndReturnResult`
have drastically changed.
And `\Rebing\GraphQL\GraphQLController::executeQuery` is gone completely.

Also, creating the "rootValue" and "contextValue" is now not baked in anymore, as an example for the context, has "just become another middleware".

I like the simplicity that the core now has been turned into this:
```php
    public function query(string $schemaName, OperationParams $operationParams, $rootValue = null, $contextValue = null): array
    {
        try {
            $result = $this->queryAndReturnResult($schemaName, $operationParams, $rootValue, $contextValue);
        } catch (Error $error) {
            $result = new ExecutionResult(null, [$error]);
        }

        return $this->decorateExecutionResult($result)->toArray();
    }

    public function queryAndReturnResult(string $schemaName, OperationParams $params, $rootValue = null, $contextValue = null): ExecutionResult
    {
        return $this->app->make(Pipeline::class)
            ->send([$schemaName, $params, $rootValue, $contextValue])
            ->through($this->executionMiddleware($schemaName))
            ->via('resolve')
            ->thenReturn();
    }
```
and everything else happens in the middlewares.

---
Now the cons are of course the massive breaking changes. It's not complete hyperbole to saying this works so differently, this is almost a different library. Anyone extending the slightest parts of this library will have to adjust things big time.

### On demand schemas: gone!
Also there's one feature now completely gone, which was supporting "on demand schemas" https://github.com/rebing/graphql-laravel/blob/1a7d5d7cdf53ea1c3db3db7e6cfdb1deb320fad9/src/GraphQL.php#L134

`$opts['schema']` actually can be an array containing `'query'` and `'mutation'` which is then used to build the schema a few lines below. This is never used that way when just accessing the whole GraphQL endpoint via controller, because the schema name is derived from the route and thus always references the schemas as configured in the configuration.

But for direct uses, like we also have it in some tests, this is gone now. I also think with the changed architecture, it's not possible to bring this back: the parameters accepted by `\Rebing\GraphQL\Support\ExecutionMiddleware\AbstractExecutionMiddleware::handle` are very specific (except what gets passed 1:1 to GraphQL, i.e. rootValue and contextValue) and there's just no room for passing custom data structures around.

I need to know some practical uses cases for this, but I think being able to use `\Rebing\GraphQL\Support\Contracts\ConfigConvertible` for dynamic schemas is actually the way to go (but it's not a real replacement).

---
Yes, I think I got super carried away. But thanks to **your** pipeline implementation, I was finally able to express and fix what I was loathing for some time already.

Let me know what you think about this.

crissi:
awesome work, looks good:)

crissi:
Let me know if you need my help:)

mfn:
Thanks :)

Currently I'm hard thinking if it's a good idea to remove the way the schema is supported in master or not, i.e. the "On demand schemas: gone"

Then hard thinking again if the internal API of the methods is the best.

Only _then_ I'm going to adapt all the tests.

_Likely_ I will want to switch the tests from the internal API to calling the route rather, as this is simply easier to manage in general IMHO…

mfn:
So I devised a trick for now: a reinstated the `query` and `queryAndReturnResult` methods almost as is:
- I marked them `@deprecated`
- since the `detectUnusedVariables` in the form it was present on the mater branch does not exist anymore, I removed
- I didn't touch them otherwise
- this I've renamed the ones which changes so drastically as `executeQuery` and `executeQueryAndReturnResult`; I know, creative 😅

In the meantime in master, I refactored the tests in a way to make them more compatible and actually less reliant on that particular `query` method and merged it into this branch.

OTOH, since I didn't really change those methods, they are **not** running any middleware ❗ 

---
- I'm not sure I want to keep it this way, but I need to get back a full working test suite to spot otherwise hidden problems; luckily seems there are none
- I'm still pondering about the signature change regarding `Schema $schema`, but that has to wait for another day => 💤

mfn:
I've further refactored the existing query/queryAndReturnResult methods so they now use the new approach with the middleware.

This (again) required some slight breaking method signatures but I think from a users perspective, e.g. using the Facade like `GraphQL::query()` this will mostly remain unchanged.

However the bigger thing is that the middleware signature now is this:
`handle(string $schemaName, Schema $schema, OperationParams $params, $rootValue, $contextValue, Closure $next)`

I.e. I added the `Schema $schema`

This actually makes `$schemaName` obsolete, at least what the current middlewares do (or rather: do not do) with: it's essentially unused.

However for now I decided to keep it, because it gives a guaranteed way to look into config via `graphql.schemas.$schemaName` in case custom additions have been added.

crissi:
All good man, the feature is the most important thing for me😊

CHANGELOG.md
```diff
@@ -94,6 +94,7 @@ CHANGELOG
       new: `public function queryAndReturnResult($query, ?array $variables = [], array $opts = []): ExecutionResult`
 
 ### Added
+- The primary execution of the GraphQL requested is now modeled via middlewares [\#762 / crissi and mfn](https://github.com/rebing/graphql-laravel/pull/762)
```
mfn: @crissi I don't want it to go unmentioned I also mentioned me as secondary author, is this fine with you?

crissi: all good :slightly_smiling_face:

crissi:
Getting "Type default not found". Checked the config, but looks similar. 

image

I am firing a query from graphiql UI. Are you getting the same?

mfn:
> Getting "Type default not found". Checked the config, but looks similar.

I've recently changed the behaviour that invalid schemas are not silently ignored anymore when building the routes => https://github.com/rebing/graphql-laravel/pull/766

- Is your config `graphql.default_schema` set to `default`?
- Do you have a `graphql.schemas.default`?

image

That would be my first idea…

crissi:
> > Getting "Type default not found". Checked the config, but looks similar.
> 
> I've recently changed the behaviour that invalid schemas are not silently ignored anymore when building the routes => #766
> 
> * Is your config `graphql.default_schema` set to `default`?
> * Do you have a `graphql.schemas.default`?
> 
> image
> 
> That would be my first idea…

thanks , error 40:)

mfn:
> do we need a getter and setter on OperationParams for the variables? now there is only the public variables

I wanted to stick to and follow `\GraphQL\Server\OperationParams` which works that way.

I'm not really happy how `\Rebing\GraphQL\Support\OperationParams` turns out regarding the decoration and might revisit this, but otherwise no real plans here.

> thanks , error 40:)

Do you mean it works for you?

crissi:
Yes works fine😊

PR #763, opened 2021-05-03
Fix example classes in config

mikefrancis:
## Summary

The example classes in the config do not use the same folders/namespaces as the artisan commands generate.

For example:

```bash
php artisan make:graphql:query SomeQuery
```

Will create the following file: `./app/GraphQL/Queries/SomeQuery.php` (notice the plural `Queries`)

In the config, the examples are all give with the singular. For example:

```php

  // ...
  'query' => [
    App\GraphQL\Query\UsersQuery::class,
  ],
```

Not a huge change but helps with consistency. 

Feedback welcomed!

---

Type of change:

Checklist:
- [ ] Update the README.md

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -45,7 +45,7 @@ return [
     //      'default' => [
     //          'controller' => MyController::class . '@method',
     //          'query' => [
-    //              App\GraphQL\Query\UsersQuery::class,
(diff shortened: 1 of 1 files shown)


mfn (approved):
Nice! Consistency FTW! 💪 

Thank you!

PR #765, opened 2021-05-06
schema: a query is always required

mfn:
## Summary
Also simplify logic on surrounding code a bit, don't call `->objectType`
if we don't have anything defined.

Thing is: this breaks a lots of test and "usually" it just works.

I've also seen in practice, a schema without a "query" just works but when you try to actually validate the schema, it reports an error.

Reference in the spec https://spec.graphql.org/June2018/#sec-Root-Operation-Types
> The query root operation type must be provided and must be an Object type.

---

Type of change:

Checklist:
- [ ] Update the README.md

### DIFF
--- a/src/GraphQL.php
+++ b/src/GraphQL.php
@@ -71,7 +71,7 @@ class GraphQL
             return $schema;
         }
 
-        $schemaQuery = $schema['query'] ?? [];
+        $schemaQuery = $schema['query'];
         $schemaMutation = $schema['mutation'] ?? [];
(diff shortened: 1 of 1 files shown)


mfn:
I've extracted this part into a separate commit into master
>  don't call ->objectType if we don't have anything defined

=> https://github.com/rebing/graphql-laravel/commit/1a4ebc6b3ac60a0790ff4b78b0431af518a758f4

But I think otherwise I'll rest this one for now.

PR #793, opened 2021-06-08
use webonyx queryplan

crissi:
## Summary
Uses webonyx query plan instead of ResolveInfoFieldsAndArguments
- This is also a pre step in fixing **SelectFields** class lacking functionality in regards to interface and union types. Since the **queryplan** has additional option to give us for info about the quering of these

I removed the depth, since I don' think it is actually needed anymore. So might need an documentation update.

Type of change:

Checklist:
- [x] Update the README.md


src/Support/SelectFields.php
```diff
@@ -192,7 +192,8 @@ protected static function handleFields(
                     );
                 }
                 // With
-                elseif (is_array($field['fields']) && $queryable) {
+
+                elseif (is_array($field['fields']) && !empty($field['fields']) && $queryable) {
```
mfn, line 196: I see that lots-o-tests are failing without the empty check, but failed to properly understand: why, now?

tests/Database/SelectFields/DepthTests/DepthTest.php
```diff
@@ -64,13 +64,9 @@ public function testDefaultDepthExceeded(): void
...
 
-        $result = $this->httpGraphql($graphql, [
-            'expectErrors' => true,
-        ]);
+        $result = $this->httpGraphql($graphql);
 
-        self::assertCount(1, $result['errors']);
-        self::assertSame('Internal server error', $result['errors'][0]['message']);
-        self::assertSame('SQLSTATE[HY000]: General error: 1 no such column: posts.user (SQL: select "posts"."id", "posts"."user", "posts"."user_id" from "posts" where "posts"."user_id" in (1) order by "posts"."id" asc)', $result['errors'][0]['debugMessage']);
+        self::assertCount(0, $result['errors'] ?? []);
```
mfn, line 69: Testing for "absence of errors" is weird with the current way the `httpGraphql` works 😬: because _if_ there's an error, it will abort the execution already and never hit any assertions.

Nevertheless, I might think that it would make it more clear to use this, WDY?
```suggestion
        self::assertArrayNotHasKey('errors', $result);
```

mfn (approved):
Wow that looks great!

> So might need an documentation update.

Yes, definitely:
- the readme references the `$getSelectFields` closure, which I guess now does not accept the depth anymore?
- it also should be added to the changelog (not just into the "Changed" section but also into the BC break one)

Can you also do that?

src/Support/SelectFields.php
crissi: it returns **fields** for simple properties also:

image


!empty might actually be enough here, will check

crissi:
> Wow that looks great!
> 
> > So might need an documentation update.
> 
> Yes, definitely:
> 
> * the readme references the `$getSelectFields` closure, which I guess now does not accept the depth anymore?
> * it also should be added to the changelog (not just into the "Changed" section but also into the BC break one)
> 
> Can you also do that?

Yes will do

mfn:
Thanks! Forgot to mention, the actual removal of the `class \Rebing\GraphQL\Support\ResolveInfoFieldsAndArguments` should also be in the BC section.

Reminds me that we should probably use `@api` or some other marker to make clear what things are considered part of the "library API" and which not, as PHP has no "package visibility"…

composer.json
```diff
@@ -37,7 +37,7 @@
         "php": ">= 7.2",
         "illuminate/contracts": "^6.0|^7.0|^8.0",
         "illuminate/support": "^6.0|^7.0|^8.0",
-        "webonyx/graphql-php": "^14.0",
+        "webonyx/graphql-php": "^14.6.4",
```
crissi, line 40: the reason why I upped this was because of a bug in the queryplan that fixed in this release

mfn:
Thank you @crissi , this really looks like a nice detail!

PR #799, opened 2021-06-24
make it easier to extend select fields

crissi:
## Summary
It is pretty hard to extend select fields at this moment, you need to copy the whole orignalResolver over. This should make it a lot easier.

Type of change:
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [x ] Misc. change (internal, infrastructure, maintenance, etc.)

Checklist:
- [ ] Existing tests have been adapted and/or new tests have been added
- [ ] Add a CHANGELOG.md entry
- [ ] Update the README.md
- [ ] Code style has been fixed via `composer fix-style`

PR #806, opened 2021-07-21
allow adding columns in query field

crissi:
## Summary
Allow adding columns in query closure. Fixes: https://github.com/rebing/graphql-laravel/issues/875

People might  be using addSelect() in their model relationship or have a need to ad an extra column to a query direcly in the custom query.

Type of change:

Checklist:
- [x ] Existing tests have been adapted and/or new tests have been added
- [x] Update the README.md


src/Support/SelectFields.php
```diff
@@ -98,12 +98,12 @@ public static function getSelectableFieldsAndRelations(
...
         return function ($query) use ($with, $select, $customQuery, $requestedFields, $ctx): void {
+            $query->select($select);
+            $query->with($with);
+            
             if ($customQuery) {
                 $query = $customQuery($requestedFields['args'], $query, $ctx);
             }
-
-            $query->select($select);
-            $query->with($with);
```
mfn, line 106: This is a BC break: previously, `$customQuery` could returned a very different/completely custom `$query` but now that's not possible anymore.

For the fullest flexibility, we would need both:
- a `$createCustomQuery` (i.e. the current functionality)
- a `$adaptQuery` (i..e. what you're proposing)

mfn (changes requested):
See my inline feedback, this is a BC break.

I can see the argument for your use case, but then we break the others and there's no "right or wrong" here. We currently support the one case and afterwards only the other.

Basically for this to work, we would need to keep the current `'query'` and add another one, invoked afterwards.

Interestingly, now that your https://github.com/rebing/graphql-laravel/pull/799 is merged, you could extend `SelectFields`, override `getSelectableFieldsAndRelations` and make any logic you want…

What do you think? How do you want to proceed here? I can't merge "as is" as I don't see a good argument made for changing/dropping the _existing_ behavuour.

src/Support/SelectFields.php
crissi: Hmm,  I might have meant to use addSelect, let me modify that and see if that changes anything

crissi:
Hi mfn. 
I opened again using, addSelect and not reordering the way with and select  works

mfn (commented):
Looks more elegant, but technically it's still change of behaviour…

Test coverage LGTM and I also can't think of case were this actually would break.

Are we sure we've no concerns here? Or should be rather wait for a major version bump?

We would also need a good / explanatory changelog for that one, so ppl understand the impact of the change.

WDYT?

crissi:
could add it to this example: https://github.com/rebing/graphql-laravel#type-relationship-query
so it also included addSelect( 

I personally dont think this can cause any errors.

mfn (commented):
> could add it to this example

Ok, let's do this. Yes, please add it :)

crissi:
> > could add it to this example
> 
> Ok, let's do this. Yes, please add it :)

added

mfn (approved):
Thank you!

PR #823, opened 2021-09-15
Add optimization

kasian-sergeev:
## Summary

Move getting name after cast to array. This removes extra call getAttributes() from Field::__get() magic method

Type of change:

Checklist:
- [ ] Update the README.md

### DIFF
--- a/src/GraphQL.php
+++ b/src/GraphQL.php
@@ -329,8 +329,7 @@ class GraphQL
                 $field = $this->app->make($field);
                 $field = $field->toArray();
             }
-            $nameFromField = $field['name'] ?? null;
-            $name = is_numeric($name) ? $nameFromField : $name;
+            $name = is_numeric($name) ? $field['name'] : $name;
             $field['name'] = $name;
             $typeFields[$name] = $field;
         }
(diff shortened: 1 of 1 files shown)


mfn (changes requested):
> This removes extra call getAttributes() from Field::__get() magic method

Do you have a benchmark what the impact of this is?

I appreciate the change, but I'm also worried about the readability of the code and thus would like to understand the performance fix here.

To illustrate what I mean: the introduction of the `null`, which wasn't there before, means mentally I see scenario where `$field['name']` gets null assigned, which wasn't there before:
image


_Practically_ it's likely not possible, but now I find the code in fact less clear on its intention.

kasian-sergeev:
I tested it with ab on my work project and found no performance improvement. But I think this behavior may affect performance when this method uses an external store (like db) for validation:
```php
public function args(): array
{
    $colorNames = Color::query()->select('name')->get()->collapse()->all();

    return [
        'colorName' => [
            'rules' => [ 'in:' . implode(',', $colorNames)],
        ],
    ];
}
```
It's a bad practice to use such validation without a closure here, and yet it's possible.

mfn (approved):
Final state of PR makes sense, thank you! 👍🏼

PR #828, opened 2021-09-29
Fix 'TypeNotFound' when an interface defined after another type where it is used

kasian-sergeev:
## Summary

Throws 'Rebing\GraphQL\Exception\TypeNotFound' when an interface defined in 'types' after another type where it is used.

To reproduce the bug put LikableInterfaceType::class under PostType::class in InterfaceTest::getEnvironmentSetUp() and run tests.

I think that the method 'interfaces' should be used in the same way as 'getFields'. A possible disadvantage of my version is that now there is always a 'interfaces' field in the array, which was not before. I did not dig deep into the webonyx library and am not sure that this will not break any functionality.

Type of change:

Checklist:
- [ ] Update the README.md

### DIFF
--- a/src/Support/Type.php
+++ b/src/Support/Type.php
@@ -97,19 +97,15 @@ abstract class Type implements TypeConvertible
     public function getAttributes(): array
     {
         $attributes = $this->attributes();
-        $interfaces = $this->interfaces();
 
-        $attributes = array_merge($this->attributes, [
+        return array_merge($this->attributes, [
             'fields' => function () {
                 return $this->getFields();
             },
+            'interfaces' => function () {
+                return $this->interfaces();
+            },
         ], $attributes);
-
-        if ($interfaces) {
-            $attributes['interfaces'] = $interfaces;
-        }
-
-        return $attributes;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


mfn (changes requested):
I like the change, it effectively makes interfaces also lazy evaluated which I assume is the trick to fix those type dependency problems you are referring to!

However before proceeding here, I'd like to request two things:
- the easy part: can you please update your PR with latest master branch? Even when you submitted it, it was already outdated by a few weeks
  I tried to fix it but I can't push to your repo
- the no-so-easy part: The test you provided is nice, but it really doesn't reflect the impact of the change regarding the resolver. I kindly ask that you come up with a test which shows the code change fixes it!

Thank you!

mfn (approved):
> To reproduce the bug put LikableInterfaceType::class under PostType::class in InterfaceTest::getEnvironmentSetUp() and run tests.

I missed that part on my first review round!

This actually is enough for me to adapt the tests so we've a reproducible.

Since I couldn't push to your branch, I kept your commit but adjusted the tests, changelog and merged with master in https://github.com/rebing/graphql-laravel/pull/847

Thank you!

PR #830, opened 2021-10-01
Add command for schema class generation

matsn0w:
## Summary
- Added a new command: `php artisan make:graphql:schema <name>`
- Added test for new command

This new command can be used to quickly generate a new schema configuration class.

Type of change:

Checklist:
- [x] Update the README.md


src/Console/SchemaMakeCommand.php
```diff
@@ -0,0 +1,42 @@
...
+        return $this->replaceGraphqlName($stub);
+    }
+
+    protected function replaceGraphqlName(string $stub): string
+    {
+        $graphqlName = lcfirst($this->getNameInput());
+        $graphqlName = preg_replace('/Schema$/', '', $graphqlName);
+
+        return str_replace(
+            'DummyGraphqlName',
```
mfn, line 37: Shouldn't this match the class `DummyClass`? 🤔

mfn (changes requested):
LGTM!

But I think the command (and this all the class / test names, etc.) should be named `schemaConfiguration` (also fine with `schemaConfig`).

Otherwise it might be confusing because a "graphql schema" is definitely something else.

WDYT?

src/Console/SchemaMakeCommand.php
matsn0w: oh lol, that entire method is a copy/paste artifact XD

will fix in a minute

matsn0w: see https://github.com/rebing/graphql-laravel/pull/830/commits/e3e905e080103d98207c3f0b40e269ddc22e9313

matsn0w:
> But I think the command (and this all the class / test names, etc.) should be named `schemaConfiguration` (also fine with `schemaConfig`).
> 
> Otherwise it might be confusing because a "graphql schema" is definitely something else.
> 
> WDYT?

That makes sense. I'd go with the shorter `schemaConfig` for the sake of not having to type that out entirely ;)

mfn (approved):
LGTM!

---
However I've to merge your PR via https://github.com/rebing/graphql-laravel/pull/840 because you had PHP 8 specific changes (the project is PHP 7.2+) and some other phpstan issue I solved there.

Rest assured your authorship is retained!

PR #843, opened 2021-11-04
Add array type hint for $args in stub files

sforward:
## Summary

The $args argument in the stub files could be type hinted as array. The return value of Rebing\GraphQL\Support\Field::getArgs is passed, which always is an array.

---

Type of change:

Checklist:
- [x] Update the README.md

### DIFF
--- a/src/Console/stubs/middleware.stub
+++ b/src/Console/stubs/middleware.stub
@@ -10,7 +10,7 @@ use Rebing\GraphQL\Support\Middleware;
 
 class DummyClass extends Middleware
 {
-    public function handle($root, $args, $context, ResolveInfo $info, Closure $next)
+    public function handle($root, array $args, $context, ResolveInfo $info, Closure $next)
     {
         return $next($root, $args, $context, $info);
     }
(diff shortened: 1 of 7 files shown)


src/Console/stubs/middleware.stub
```diff
@@ -10,7 +10,7 @@ use Rebing\GraphQL\Support\Middleware;
 
 class DummyClass extends Middleware
 {
-    public function handle($root, $args, $context, ResolveInfo $info, Closure $next)
+    public function handle($root, array $args, $context, ResolveInfo $info, Closure $next)
```
mfn, line 13: This one is slightly problematic, because this will _override_ the parent (in all the other stub-casses, there's actually no parent method, so it's definitely fine).

If we want to keep it here, we:
- need to update the actual Middleware
- mention it in the changelog
  Though I don't thin it's a breaking change, it's a change we should mention

mfn (changes requested):
Heya, thanks for the PR!

LGTM except "this one thing"

Also, there are a _lot_ of mentions in the readme reagarding `$args`: would you think it also make sense changing it there as a "good practice"?

src/Console/stubs/middleware.stub
sforward: Good catch, I've changed it in \Rebing\GraphQL\Support\Middleware as well and marked it breaking.

mfn (approved):
LGTM!

There's was one mire fix necessary, see https://github.com/rebing/graphql-laravel/commit/97a778acacedbbb75827fe3896bb0cd239439c16

PR #856, opened 2021-11-13
Don't require a return value for the query option of fields

sforward:
## Summary

The field's `query` option acts like Laravel's query builder scope, but unlike a scope it currently requires the query builder as return value. If omitted (as I've done several times out of habit), you get a non-descriptive error: "Call to a member function select() on null". With this PR the query builder object will just be used if none was returned.

---

Type of change:

Checklist:
- [x] Update the README.md


mfn (approved):
Oh, that's actually very nice!

mfn:
PS: updated changelog via https://github.com/rebing/graphql-laravel/commit/5817033018161eeab1092d2ae7293fded4bdebbd

PR #867, opened 2021-12-11
Update README.md

leonardocustodio:
## Summary
I fixed a logic error in this part of the doc. The args comes from $args not from function($inputArguments) as this will return the key value.

Checklist:
- [X] Update the README.md


mfn (approved):
🙄 

Thank you! 🙏🏼

Issue #875, opened 2021-12-31
Can't control loaded columns/relations in custom query


codercms:
### Versions:
- graphql-laravel Version: 8.0.0
- Laravel Version: 8.x
- PHP Version: 8.0

### Description:
In case when need to load dependencies for non selectable field for eager loaded relation I try to load them in custom query:
```php

'positions' => [
    'type' => Type::nonNull(Type::listOf(Type::nonNull(GraphQL::type('Position')))),
    'query' => static function (array $args, BelongsToMany $query, Context $ctx) {
        /** @var ResolveInfo $info */
        $info = $ctx->get('info');
        if ($info !== null) {
            $selectedFields = $info->getFieldSelection(2)['positions'];
            if ($selectedFields['virtual_field_which_depend_on_other_fields'] ?? false) {
                   $query->addSelect(['dependcy1', 'dependency2']);
            }
        }
    },
]
```

But my `addSelect` statement is immediately overwritten in this block of code -  https://github.com/rebing/graphql-laravel/blob/master/src/Support/SelectFields.php#L105:
```php
        return function ($query) use ($with, $select, $customQuery, $requestedFields, $ctx): void {
            if ($customQuery) {
                $query = $customQuery($requestedFields['args'], $query, $ctx) ?? $query;
            }

            $query->select($select);
            $query->with($with);
        };
```
I think it's a quite buggy behavior, because custom query should have full control on which fields/relations will be loaded.
Maybe this fragment of code should look like:
```php
        return function ($query) use ($with, $select, $customQuery, $requestedFields, $ctx): void {
            $query->select($select);
            $query->with($with);

            if ($customQuery) {
                $query = $customQuery($requestedFields['args'], $query, $ctx, /** how about to pass also $select and $with parameters? *//) ?? $query;
            }
        };
```

### Steps To Reproduce:

- …

crissi:
https://github.com/rebing/graphql-laravel/pull/806 i made pr that we might take a look at again too see if we can make it possible to suport that

Issue #510, opened 2019-10-29
Allow manual adding to $getSelectFields


georgeboot:
## Summary
It would be very powerful if developers can manually add fields, relations and counts to the `$getSelectFields`.

### Background
Imagine the following types:
```gql
type Event {
    id: ID!
    name: String
    tickets: [Ticket]
}

type Ticket {
    id: ID!
    event: Event!
    date: Date
    name: String
    price: Int
}
```

As you can see, we have an event with multiple tickets. Each ticket has a date. These dates don't have to be unique.

Lets imagine we have the following tickets:

| Name  | Date | Tickets Available |
| ------------- | ------------- | --: |
| Day 1 Regular  | 2020-01-01 | 100 |
| Day 1 VIP  | 2020-01-01 | 10 |
| Day 2 Regular  | 2020-02-01 | 100 |
| Day 2 VIP  | 2020-01-02 | 10 |
| Day 3 Regular  | 2020-03-01 | 100 |
| Day 3 VIP  | 2020-01-03 | 10 |

Now, from the event I want to have a list of available dates.

```gql
type EventDay {
    date: Date
    ticketsAvailable: Int
}
type Event {
    id: ID!
    name: String
    tickets: [Ticket]
    dates: [EventDay]
}
```

The expected output:
```json5
{
  "data": {
    "event": {
      "id": "1",
      "dates":[
        {
          "date": "2020-01-01",
          "ticketsAvailalbe": 20 // we already have 90 total sales for this date
        },
        {
          "date": "2020-01-01",
          "ticketsAvailalbe": 100 // total of 10 sales
        },
        {
          "date": "2020-01-01",
          "ticketsAvailalbe": 110 // no sales yet
        }
      ]
    }
  }
}
```

### The problem
I've added a `getDatesAttribute` method to my `Event` model. In there, I select all tickets, filter out some unwanted statuses, and for every ticket, we get the sales count (the `Tickets Available` from the example table is calculated from our `sales` table) et cetera.

There is unfortunately no way (that I know) to tell the `Event` resolver to eager load some things so that the `getDatesAttribute` will not create a `n+1` issue.

### My idea
It would be very powerful if I could specify what fields to eager load in my `App\GraphQL\Types\EventDay`.

Also, the `$getSelectFields` currently only supports fields and relations, not counts. I would also love to see that one added.

With these improvements, my `EventDay` type will inform my `Event` type that it would love to eager load things like `$event->loadCount('tickets.sales')` etc.

[...]

PR #877, opened 2022-01-02
Make it possible to modify eager-loading queries by extending SelectFields

turgutsaricam:
## Summary
This project is a tremendous help. However, when it comes to eagerly loading relations, I always end up in examining `SelectFields` class to figure out how to implement the functionality I want. Most of the time, I simply need to modify the eager-loading query, created in `SelectFields` class, by considering the requested fields. I probably spent more than 100 hours in total trying to hack the package to implement what I need. The reason I spent this much time is because `SelectFields` class is not extendible. For example, #510 talks about eagerly loading counts, which is not possible currently. The reason why I created this pull request is that I need to eagerly load counts as well. There are also other issues related to `SelectFields` in this repository. For example, #875 mentions a need for a custom behavior. #806 wants to modify the behavior of eager-loading function. #799 was a nice attempt, which made it possible to extend `SelectFields` and provide a custom `SelectFields` class to be used by the package, but it is not enough for many use-cases. The class is quite complex, as also stated [here](https://github.com/rebing/graphql-laravel/pull/799#discussion_r657792900). The complexity of the class partially comes from using static methods too. This pull request partially addresses these issues.

Here is a summary of the changes made in this pull request:
- Use instance methods instead of static methods. By this way, we do not need to pass certain variables as parameters to all the methods.
- Store constructor parameters in instance variables. Their getters are added as well. With this, a child class can retrieve the constructor parameters from any method, when overriding the methods. Also, we do not need to pass `$queryArgs` and `$ctx` variables. The methods that need them simply retrieve them via the getters. This reduces the complexity of the class.
- Make it possible to modify the eager-loading queries. This is the main contribution of this pull request.  This will make it possible to customize the behavior of eager-loading.
- Do not override the columns added to the select statement of the queries by the custom queries. This lets the custom queries define custom columns.

There are many [eager-loading methods in Laravel](https://laravel.com/docs/8.x/eloquent-relationships#other-aggregate-functions). We might also need to customize the eager-loading behavior to use them as well. With the changes in this pull request, everybody can implement their own behavior. Below is an example of extending `SelectFields` to eagerly load counts:

```php
<?php

namespace App\GraphQL\Helpers\SelectFields;

use App\GraphQL\Fields\Pagination\CustomPaginationType;
use GraphQL\Type\Definition\Type as GraphqlType;
use Illuminate\Database\Eloquent\Relations\Relation;
use Illuminate\Support\Str;
use Rebing\GraphQL\Support\SelectFields;

class AppSelectFields extends SelectFields {

    // Override the onAfterModifyQuery method to modify the query after select fields and relations 
    // are added to it. Also, by overriding onBeforeModifyQuery, it is possible to modify $with and
    // $select variables that will be added to the query in SelectFields.
    protected function onAfterModifyQuery($query, GraphqlType $parentType, array $with, array $select,
                                          array $requestedFields): void {
        parent::onAfterModifyQuery($query, $parentType, $with, $select, $requestedFields);

        if ($parentType instanceof CustomPaginationType) {
            $type = $parentType->getUnderlyingType();
            $actualRequestedFields = data_get($requestedFields, 'fields.items.fields') ?? [];
            $this->eagerlyLoadCounts($query, array_keys($actualRequestedFields), $type);
        }
    }

    /*
     *
     */

    /**
     * @param Relation    $query
     * @param string[]    $requestedFields
     * @param GraphqlType $type
     * @return void
     * @noinspection PhpMissingParamTypeInspection
     */
    protected function eagerlyLoadCounts($query, array $requestedFields, GraphqlType $type) {
        foreach($requestedFields as $fieldName) {
            if (!str_ends_with($fieldName, '_count')) continue;

            $exploded = explode('.', $fieldName);
            $this->eagerlyLoadCount($query, $exploded[count($exploded) - 1], $type);
        }
    }

    /**
     * @param Relation    $query
     * @param string      $fieldName
     * @param GraphqlType $type
     * @return void
     */
    protected function eagerlyLoadCount($query, string $fieldName, GraphqlType $type): void {
        $field = $type->findField($fieldName);
        if (!$field) return;

        $config = $field->config;
        // 'count' is a custom config that I added to customize the behavior
        if (($config['count'] ?? true) !== true) return;

        // 'relation' is a custom config that I added to customize the behavior
        $relationName = $config['relation'] ?? Str::replace('_count', '', $fieldName);
        $query->withCount($relationName);
    }

}
```

Switching to instance methods from static methods opens up the possibility of adding more hooks like `onBeforeModifyQuery` and  `onAfterModifyQuery`. In the future, I expect other pull requests that add other hooks when developers come across other use-cases.

The changes I made are breaking changes for the applications that extend `SelectFields`. Other than that, this is non-breaking. I did not add any tests. The existing tests all pass. If this pull request is intended to be merged, I can add the tests with your directions. I did not update `README.md` because I did not see any guidance on how to extend `SelectFields`.

---

Type of change:

Checklist:
- [ ] Update the README.md

### DIFF
--- a/src/Support/SelectFields.php
+++ b/src/Support/SelectFields.php
@@ -67,7 +67,7 @@ class SelectFields
         ];
 
         /** @var array{0:mixed[],1:mixed[]} $result */
-        $result = $this->getSelectableFieldsAndRelations($requestedFields, $parentType, null, true, $ctx);
+        $result = $this->getSelectableFieldsAndRelations($requestedFields, $parentType);
 
         [$this->select, $this->relations] = $result;
     }
@@ -76,7 +76,6 @@ class SelectFields
      * Retrieve the fields (top level) and relations that
      * will be selected with the query.
      *
-     * @param mixed $ctx The GraphQL context; can be anything and is only passed through
      * @return array|Closure - if first recursion, return an array,
      *                       where the first key is 'select' array and second is 'with' array.
      *                       On other recursions return a closure that will be used in with
@@ -85,8 +84,7 @@ class SelectFields
         array $requestedFields,
         GraphqlType $parentType,
         ?Closure $customQuery = null,
-        bool $topLevel = true,
-        $ctx = null
+        bool $topLevel = true
     ) {
         $select = [];
         $with = [];
@@ -97,7 +95,7 @@ class SelectFields
         $parentTable = $this->getTableNameFromParentType($parentType);
         $primaryKey = $this->getPrimaryKeyFromParentType($parentType);
 
-        $this->handleFields($requestedFields, $parentType, $select, $with, $ctx);
+        $this->handleFields($requestedFields, $parentType, $select, $with);
 
         // If a primary key is given, but not in the selects, add it
         if (null !== $primaryKey) {
@@ -112,9 +110,9 @@ class SelectFields
             return [$select, $with];
         }
 
-        return function ($query) use ($with, $select, $customQuery, $requestedFields, $ctx): void {
+        return function ($query) use ($with, $select, $customQuery, $requestedFields): void {
             if ($customQuery) {
-                $query = $customQuery($requestedFields['args'], $query, $ctx) ?? $query;
+                $query = $customQuery($requestedFields['args'], $query, $this->getCtx()) ?? $query;
             }
 
             $query->select($select);
@@ -139,14 +137,12 @@ class SelectFields
      * @param array<string,mixed> $requestedFields
      * @param array $select Passed by reference, adds further fields to select
      * @param array $with Passed by reference, adds further relations
-     * @param mixed $ctx The GraphQL context; can be anything and is only passed through
      */
     protected function handleFields(
         array $requestedFields,
         GraphqlType $parentType,
         array &$select,
-        array &$with,
-        $ctx
+        array &$with
     ): void {
         $parentTable = $this->isMongodbInstance($parentType) ? null : $this->getTableNameFromParentType($parentType);
 
@@ -181,7 +177,7 @@ class SelectFields
             }
 
             // First check if the field is even accessible
-            $canSelect = $this->validateField($fieldObject, $ctx);
+            $canSelect = $this->validateField($fieldObject);
 
             if (true === $canSelect) {
                 // Add a query, if it exists
@@ -199,8 +195,7 @@ class SelectFields
                         $field,
                         $fieldType->getWrappedType(),
                         $select,
-                        $with,
-                        $ctx
+                        $with
                     );
                 }
                 // With
@@ -223,8 +218,7 @@ class SelectFields
                             $field,
                             $newParentType,
                             $customQuery,
-                            false,
-                            $ctx
+                            false
                         );
                     } elseif (is_a($parentTypeUnwrapped, \GraphQL\Type\Definition\InterfaceType::class)) {
                         $this->handleInterfaceFields(
@@ -232,13 +226,12 @@ class SelectFields
                             $parentTypeUnwrapped,
                             $select,
                             $with,
-                            $ctx,
                             $fieldObject,
                             $key,
                             $customQuery
                         );
                     } else {
-                        $this->handleFields($field, $fieldObject->config['type'], $select, $with, $ctx);
+                        $this->handleFields($field, $fieldObject->config['type'], $select, $with);
                     }
                 }
                 // Select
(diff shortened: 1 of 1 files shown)


mfn:
I've spoken out multiple times against the use of `SelectFields` due to all the problems it has and you touched quite nicely upon this topic and probably more.

I exclusively use the [dataloader approach](https://www.google.com/search?hl=en&q=graphql%20dataloader%20php) and never use(d) `SelectFields` and never will. I think it's an battle/approach which cannot really be won and/or requires too much time to get right and maintain, which I'm not able to provide.

Your approach might be better, make things correct, better for the future but I just need to look at a few lines of the PR and realize it's major breaking change and would require bump of a new major version; like right after I realized 8.0.0 we would need to jump to 9.0.0 alone due to this PR 😅

---

I'm trying to navigate the muddy waters of this part of the library and find a safe harbor and I bet you invested already more time then I ever did into it.

I wonder if `SelectFields` can be made independent in a way it would be decoupled from this library without too much effort while still being backwards compatible but also be replaceable with any other implementation.

crissi:
I use the selectfields approach for 95% of use cases and it makes very fast and simple to write most queries.

But I do agree it is a bit coupled. 

One way of decoupling it could be add all select options under one key, so instead of alias, always, is_relation being it own "root" key, we would add it to, maybe also make it possible to be set as an DTO (so we validate the params are correct), or maybe transfed to DTO. So you can't set wrong values, this could of course be done with all params.

'posts' => [
      'type'          => Type::listOf(GraphQL::type('Post')),
      'description'   => 'A list of posts written by the user',
      // Now this will simply request the "posts" column, and it won't
      // query for all the underlying columns in the "post" object
      // The value defaults to true
     'selectable' => [
          'is_relation' => false,
          'always' => ['key']
    ],
  ]

Then we could start deprecating the keys in the root.

 I would be up for making a pr for this!

turgutsaricam:
@mfn Sorry that I did not reply earlier. After you pointed out the data loader approach, I started checking it out and implementing it into my project. I see how it is a better approach, given that it was originated at Facebook. I actually started to move away from `SelectFields` after reading your comment. Actually, I could have implemented the data loader pattern earlier, if this repository encouraged the usage of that pattern instead of `SelectFields`. This repository is too intertwined with `SelectFields`. Upon reading `README.md`, anybody who starts to use this repository will be incentivized to use `SelectFields`. There is not enough guidance to using the data loader pattern. Moreover, one should do quite a bit of research to understand that pattern and use it in their projects. For example, [overblog/dataloader-php](https://github.com/overblog/dataloader-php/) does not work with [webonyx/graphql-php](https://github.com/webonyx/graphql-php/) currently. I tried to use it, but ended up with fatal errors. After that, I decided to write my own data loader implementation that works nicely with `webonyx/graphql-php`. While doing that, I did a lot more research to understand how to implement the pattern. My point is, there is not a straightforward way to use the data loader pattern with this repository. Moreover, the `README.md` is full of `SelectField` references. Many methods come with a parameter that is used to retrieve a `SelectFields` instance. As @crissi said, I use `SelectFields` approach in about 95% of the cases as well. This is because this repository encourages the use of it. Hence, if `SelectFields` is not the right way, it should be discouraged. If it is not discouraged, new users of this library will use `SelectFields` instead of the data loader pattern, which will make it much harder to move away from `SelectFields`. In my opinion, if `SelectFields` is not the right way, its usage should be discouraged, and more guidance on how to use the data loader pattern should be in `README.md`.

Apart from these, if you agree, I could make another PR that does not break backwards compatibility yet allows to override the following eager-loading function:
```php
  return function ($query) use ($with, $select, $customQuery, $requestedFields, $ctx): void {
      if ($customQuery) {
          $query = $customQuery($requestedFields['args'], $query, $ctx) ?? $query;
      }

      $query->select($select);
      $query->with($with);
  };
```

Basically, moving the creation of this function to its own protected static method will do the job for me and probably many others. I am currently overriding the entire [SelectFields::getSelectableFieldsAndRelations()](https://github.com/rebing/graphql-laravel/blob/45ce1be9864b2490f12c70dec72d107520e81e20/src/Support/SelectFields.php#L68) method just to override the eager-loading function returned from that method. Overriding just the returned function would make it more maintainable.

mfn:
> Is there currently any progress on this?

Not from my side currently. Working on anything `SelectFields` related is not something which motivates me (except maybe: moving it out so it can be independently worked on).

turgutsaricam:
I completely moved away from `SelectFields` to the data loader pattern. So, I do not work on `SelectFields` anymore.

Artem-Schander:
Hm.. too bad. I thought I could avoid to refactor our project.

mfn:
> I completely moved away from `SelectFields` to the data loader pattern. So, I do not work on `SelectFields` anymore.

Closing this PR therefore; it can always by picked up by someone motivated 🤞🏼

Issue #872, opened 2021-12-27
NonNull type for data in pagination and NonNull ListOf type


do6po:
### Versions:
- graphql-laravel Version: all with current
- Laravel Version: 8.77.1
- PHP Version:8.0

### Description:

I want make the list (data) and type inside the paginator to be required. But there is a problem with the possible selectable fields. This is due to the lack of a recursion parameter when calling the getWrappedType method in the SelectFields class on line 188. If you pass an argument to the getWrappedType method to retrieve data recursively, everything will work correctly. 

### Steps To Reproduce:

```
    public function type(): Type
    {
        return GraphQL::paginate(
            Type::nonNull(
                GraphQL::type('someType')
            )
        );
    }
```

and inside extends PaginationType

```
    protected function getPaginationFields(string $typeName): array
    {
        return [
            'data' => [
                'type' => Type::nonNull(
                    Type::listOf(
                        GraphQL::type($typeName)
                    )
                ),
...
```

mfn:
I'm not using the paginator in my projects. Do you have a suggestion how to fix this, preferable with tests?

do6po:
> I'm not using the paginator in my projects. Do you have a suggestion how to fix this, preferable with tests?

https://github.com/rebing/graphql-laravel/pull/878

jacobdekeizer:
I have the same problem, would be nice if that PR could be merged :smile:

do6po:
Hi. I am from Ukraine. We have some a little troubles now. When we will got
a clear sky on our head, i fix all troubles with PR and you can merge it.
Cheers.

пт, 4 мар. 2022 г., 20:55 jacobdekeizer ***@***.***>:

do6po:
> I have the same problem, would be nice if that PR could be merged smile

Hi. I am from Ukraine. We have some a little troubles now. When we will got
a clear sky on our head, i fix all troubles with PR and merge it.
Cheers.

lamtranb:
> Hi. I am from Ukraine. We have some a little troubles now. When we will got a clear sky on our head, i fix all troubles with PR and merge it. Cheers.

Hope the best for you.

Could we have any update on this ?

lamtranb:
> > Could we have any update on this ?
> 
> Not sure if you mean on this issue or …
> 
> A PR was provided but there where issues, so it was rejected. Other than that, we're waiting on OP or someone else to provide a fix.

Sorry for late response. I just have the same problem.
[...]

PR #878, opened 2022-01-15
Fix for non-null data list inside pagination type;

do6po:
## Summary
Fix for non-null data list inside pagination type;
- …
issue: https://github.com/rebing/graphql-laravel/issues/872
---

### DIFF
--- a/src/GraphQL.php
+++ b/src/GraphQL.php
@@ -446,7 +446,7 @@ class GraphQL
 
     public function paginate(string $typeName, string $customName = null): Type
     {
-        $name = $customName ?: $typeName . 'Pagination';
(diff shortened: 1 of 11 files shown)


mfn (changes requested):
On paper the changes look fine to me.

But the added tests don't fully cover the changes:
- no test fails when I remove the `true` parameter in `handleFields`
- no test fails when I revert the changes in `paginate`

Basically, only the change in `simplePaginate` is covered.

Can you take care of the rest? If the tests are green, please also add a changelog entry, thanks!

mfn:
@do6po are you planning on following up? Otherwise I intend to close the PR.

PR #934, opened 2022-07-26
Fix the `methods` to required is `strtoupper` string

michaelnabil230:


mfn:
I don't think it's necessary, it's documented in the changelog under "Breaking changes" that they need to be upper case https://github.com/rebing/graphql-laravel/blob/11f470c2c4f070deb0c613d8ad117133281ee1f6/CHANGELOG.md?plain=1#L99-L100

and the example config has them in upper case, too 
https://github.com/rebing/graphql-laravel/blob/054847e7d1c21d5f329d75ee973624895534efbd/config/config.php#L91-L92

michaelnabil230:
Yes, I understand that, but so that there are no problems again and we prevent the developers from not making mistakes in any case.

PR #703, opened 2020-12-07
Basic directives support

illambo:
## Summary
Attempted integration of directives support, see [changes](https://github.com/rebing/graphql-laravel/pull/703/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R2156) in readme for use case.

#### TODO
 - [ ] Directive location:
   * Request definitions:
      + [ ] query -> seems accessible via ResolveInfo  operation->directives
      + [ ] mutation -> seems accessible via ResolveInfo  operation->directives
      + [ ] subscription -> seems accessible via ResolveInfo  operation->directives
      + [x] field
      + [ ] fragment_definition
      + [ ] fragment_spread
      + [ ] inline_fragment
      + [ ] variable_definition
   * Type system definitions:
      + [ ] schema
      + [ ] scalar
      + [ ] object
      + [ ] field_definition
      + [ ] argument_definition
      + [ ] interface
      + [ ] union
      + [ ] enum
      + [ ] enum_value
      + [ ] input_object
      + [ ] input_field_definition 
 - [ ] Discuss if treating GraphQL directives as middleware or (better) pipeline (see [interesting article](https://blog.logrocket.com/treating-graphql-directives-as-middleware/)) maybe in [default field resolver](https://github.com/rebing/graphql-laravel#default-field-resolver)
 - [ ] Understand possible adaptation of [type-modifiers](https://github.com/rebing/graphql-laravel#type-modifiers)
 - [ ] Tests
 - [ ] Unlock friendsofphp/php-cs-fixer composer dependency

#### DEV NOTES
 - https://www.apollographql.com/docs/apollo-server/schema/creating-directives/
 - underlying layer seems "unready" webonyx/graphql-php#748 webonyx/graphql-php#299
 - webonyx example custom directive - https://github.com/webonyx/graphql-php/issues/308#issuecomment-431197515
 - check implementation https://github.com/diasfs/graphql-php-resolvers
 - https://spec.graphql.org/June2018/#sec-Type-System.Directives

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

### Checklist
- [x] Existing tests have been adapted and/or new tests have been added
- [x] Add a CHANGELOG.md entry
- [x] Update the README.md
- [x] Code style has been fixed via `composer fix-style`

PR #947, opened 2022-09-26
Register directives via schema config

sforward:
## Summary
With this PR you could register directives. This opens the option to explore the possibilities of directives.

Do you want me to add this to the readme? Subscriptions can be registered the same way, but aren't mentioned in de readme.

---

Type of change:

Checklist:
- [ ] Update the README.md


mfn:
We had multiple attempts in the past ( https://github.com/rebing/graphql-laravel/pull/703 , https://github.com/rebing/graphql-laravel/pull/860 ).

What is the end goal here? I think I'd be more confident knowing where this goes, as there were reasons why the previous approaches didn't pan out.

sforward:
My previous attempt was more focussed on a general support for directives. The current PR makes it just possible to register directives to the schema, so anyone can add specific functionality via a custom directive.

For example, I was attempting an execution middleware that checks for a directive on the query level to apply some specific caching. But also directives on the field level like added to the tests in #860 could be implemented with this PR, if someone needs it in an application.

mfn:
Would it be possible to provide a small / non-useless directive as an example, which we then further could also use in a test?

> Do you want me to add this to the readme?

If we can make the above work, I'd be for this!

sforward:
I've added a test for the directive registration. It includes an example directive with all available options.

PR #949, opened 2022-09-27
GraphiQL updated to the latest version

asanovr:
## Summary
GraphiQL updated to the latest version.

---

Type of change:

Checklist:
- [ ] Update the README.md

### DIFF
--- a/resources/views/graphiql.php
+++ b/resources/views/graphiql.php
@@ -6,149 +6,127 @@
  *  LICENSE file in the root directory of this source tree.
 -->
 <!DOCTYPE html>
-<html>
-  <head>
+<html lang="en">
+<head>
+    <meta charset="utf-8" />
+    <meta name="robots" content="noindex" />
+    <meta name="referrer" content="origin" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
     <title><?php echo $schema ? "$schema | " : ''; ?>GraphiQL</title>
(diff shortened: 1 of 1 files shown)


resources/views/graphiql.php
```diff
@@ -6,149 +6,127 @@
...
+    <link type="text/css" href="//unpkg.com/graphiql/graphiql.min.css" rel="stylesheet" />
+</head>
+<body>
+<div id="splash">
+    Loading&hellip;
+</div>
+<script src="//cdn.jsdelivr.net/es6-promise/4.0.5/es6-promise.auto.min.js"></script>
+<script src="https://cdn.jsdelivr.net/npm/react/umd/react.production.min.js"></script>
+<script src="https://cdn.jsdelivr.net/npm/react-dom/umd/react-dom.production.min.js"></script>
+<script src="//unpkg.com/graphiql/graphiql.min.js"></script>
```
mfn, line 42: Shouldn't all of them be version to not unexpectedly break in the future if any of them gets updated?

mfn (changes requested):
Left a comment about the missing version pinning, makes me uncomfortable without it, WDYT?

--
I wish we could get rid of the external dependencies completely TBH, but I've no good idea how, I certainly wouldn't want to host them in this repo either.

Maybe a standaline drop-in package compatible with this package would be the right way.

Issue #948, opened 2022-09-26
Add support for thecodingmachine/safe 2.0


benrowe-chuffed:
### Summary

Please consider adding support for `thecodingmachine/safe:^2.0`.

Currently this package only supports ^1.3. Not only has 1.3 not been updated in two years, it's also missing support for PHP 8.

It's possible to support both since the 2.0 interfaces are backwards compatible AFAIK.

This can be done simply by changing the composer dependency to:

```
"require": {
  "thecodingmachine/safe": "^1.1 || ^2",
  ...
}
```

mfn:
LGTM, happy to accept a PR!

PR #961, opened 2022-12-29
Add support for thecodingmachine/safe 2.0 #948

tranvantri:
#948

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
         "illuminate/contracts": "^6.0|^8.0|^9.0",
         "illuminate/support": "^6.0|^8.0|^9.0",
         "laragraph/utils": "^1",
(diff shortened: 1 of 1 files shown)


tranvantri:
Hi @mfn,

I added a function to check the compatible ability to use thecodingmachine/safe and checked it on PHP 7.4 and 8.1.

Could you please review this PR?

Many thanks.

tranvantri:
Hi @mfn,

I apologize for my inconvenience. Could you please review this PR?

Many thanks.

tranvantri:
> Nice, seems to work: all tests are green now!
> 
> Not fond of the change itself, but I guess there's no other way than a runtime check?

If we support both php version 7 and 8, then I think we can only check it on runtime. Another way is we will copy the \Safe\uksort into our Helpers and let us handle ourself.

PR #1010, opened 2023-03-24
Change closure to arrow function in `PaginationType`

milwad-dev:
## Summary

Is more readable and cleaner.


milwad-dev:
@mfn 
In my opinion, it is not mandatory that the whole package of use arrow function!
In Laravel-Framework, there is a part of the framework using Arrow-function and Closure!
Use the arrow-function when you have only one return.

milwad-dev:
@mfn 
In my opinion, I don't think is good to add arrow-function to workflows, because sometimes need to use closure.

PR #1012, opened 2023-03-24
Add support for non null type in pagination type

### LINKED ISSUE #872: NonNull type for data in pagination and NonNull ListOf type

jacobdekeizer:
## Summary

- Add support for non nullable data in the pagination type.
- Fixes #872
- Based on a previous PR by @do6po: #878
---

Type of change:

Checklist:
- [ ] Update the README.md

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -971,6 +971,46 @@ parameters:
 			path: tests/Support/Objects/ExamplesQuery.php
 
 		-
+			message: "#^Method Rebing\\\\GraphQL\\\\Tests\\\\Support\\\\Queries\\\\PostNonNullPaginationQuery\\:\\:resolve\\(\\) has no return type specified\\.$#"
+			count: 1
+			path: tests/Support/Queries/PostNonNullPaginationQuery.php
+
+		-
+			message: "#^Method Rebing\\\\GraphQL\\\\Tests\\\\Support\\\\Queries\\\\PostNonNullPaginationQuery\\:\\:resolve\\(\\) has parameter \\$args with no type specified\\.$#"
+			count: 1
+			path: tests/Support/Queries/PostNonNullPaginationQuery.php
+
+		-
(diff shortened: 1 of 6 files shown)


sforward:
Maybe a better solution is to make the type of the `data` field in `PaginationType::getPaginationFields` non-null?

jacobdekeizer:
@sforward Yes, but in theory that is a breaking change. Maybe we can change that as default for the next major release. I've currently done the same in my own project by changing the `'pagination_type'` in the graphql config file to my own pagination type with non null values.

Regardless of changing the pagination type or by providing the non null types yourself we still have to use `$fieldType->getWrappedType(true)` in the `SelectFields` class, otherwise `getSelectFields()->getSelect()` in the resolve method doesn't work correctly (which is covered by the tests in this pr).

99linesofcode:
@jacobdekeizer Since the type will become stricter I don't really see how this will break anything downstream.

Upstream, Laravel's Paginator returns a collection which will always be of type array. Never null. That has been the case since the `getCollection()` was introduced. In light of that, the return type of `PaginationType` is simply incorrect.

PR #1067, opened 2023-07-07
Update GraphQLController.php

alancolant:
Correction find schema when prefix contains parameters

## Summary

- …

---

Type of change:

Checklist:
- [ ] Update the README.md

### DIFF
--- a/src/GraphQLController.php
+++ b/src/GraphQLController.php
@@ -76,7 +76,7 @@ class GraphQLController extends Controller
 
     protected function findSchemaNameInRequest(Request $request, string $routePrefix): ?string
     {
-        $path = $request->getPathInfo();
+        $path = "/" . $request->route()->uri();
 
         if (!Str::startsWith($path, $routePrefix)) {
             return null;


alancolant:
Ok, let's imagine I want a dynamic URL parameter at my Graphql entry point.

I'll start by setting the prefix at the config level to "graphql/{my_parameter}" like this:

```php
return [
    ...
    'route'      => [
        // The prefix for routes; do NOT use a leading slash!
        'prefix'           => 'graphql/{my_parameter}',
        ...
    ],
    ...
]
```

The route generated will therefore make it possible to retrieve this parameter in this way in my queries:
```php
$myParameter = request()->route()->parameter('my_parameter')
```

This works for the current version's default schema.

But in case I have several schemas, a "public" and a "private".

The controller tries to detect the schema using 
```php
if (!Str::startsWith($path, $routePrefix)) {
    return null;
}
```

In case the URL is "/graphql/parameter_arbitrary_value/private"

```php

// When $path is retrieved using $request->getPathInfo(), the condition become

if (!Str::startsWith('/graphql/parameter_arbitrary_value/private', '/graphql/{my_parameter}')) { 
    return null; // Always return null
}


//But, when $path is retrieved using $request->route()->uri(), the condition become
if (!Str::startsWith('/graphql/{my_parameter}/private', '/graphql/{my_parameter}')) { 
    return null;
}

//Everything works fine
```

alancolant:
Here, 2 tests have been added, one for the default schema and a second for the custom schema.

alancolant:
I think a better approach would be to declare one and only one graphql route:

`$prefix/{schemaName?}`

And in your controller use

```php
$schemaName = $request->route()->parameter('schemaName',$config->get('graphql.default_schema', 'default'));
//Instead of:
$routePrefix = $config->get('graphql.route.prefix', 'graphql');
$schemaName = $this->findSchemaNameInRequest($request, "/$routePrefix") ?: $config->get('graphql.default_schema', 'default');
```

But this is probably going to be a breaking change, 
since there will no longer be the possibility to retrieve routes named 'graphql.custom' using `route('graphql.custom')`, but it will have to use `route('graphql',['schemaName'=>'custom'])` instead

alancolant:
Alright I am fine with it.
I will on my side try to implement the suggestion above.
I will just fix the problem  you have detected  for the moment.
Good holidays

alancolant:
**Comment:**

To avoid breaking changes, I have reworked the `routes.php` file so that the specified routes remain independent: `graphql`, `graphql.custom`, `graphql.default`.
However, instead of using a parameter at the route level, I have opted to automatically instantiate a middleware that injects the `schemaName` variable into `$request->server` for each route.

I find this approach more flexible as it allows us to eventually inject other information that can be retrieved in any controller using `$request->server('graphql.*')`.

I also took the opportunity to declare and move the route declaration into the `GraphQL.php` class within a `parseRoute` function.

All these changes have been tested, and some tests have been modified to verify the presence of the `GraphQLHttpMiddleware` for each generated route.

I have also removed the `findSchemaNameInRequest` function from the controller since we no longer need to detect this name from the URL.

Looking forward to your feedback on this proposal.

PR #1091, opened 2023-09-10
Implement lazy root types

sforward:
## Summary
This PR implements the lazy root types, supported since 15.6.0 of webonyx/graphql-php. I see some performance improvements by this change, tested in a project with several queries and mutations.

---

Type of change:

Checklist:
- [ ] Update the README.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ CHANGELOG
 
 [Next release](https://github.com/rebing/graphql-laravel/compare/9.1.0...master)
 --------------
+### Changed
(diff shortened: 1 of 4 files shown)


sforward:
When executing a gql query, the mutations are no longer initialized with this change, but other query classes still are. So #1087 isn't fully solved, but I think there is more possible.

sforward:
To resolve a query, all the fields of the Query root type are loaded in GraphQL::buildObjectTypeFromFields. It's possible to make them lazy as well, but only if the names are already set in de config, see #1092.

PR #1033, opened 2023-05-12
pagination: enforce `nonNull` on the inner type as well as on the list

mfn:
## Summary
- Based on https://github.com/rebing/graphql-laravel/pull/1012
- Fixes https://github.com/rebing/graphql-laravel/issues/872

---

Type of change:
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update
- [ ] Misc. change (internal, infrastructure, maintenance, etc.)

Checklist:
- [x] Existing tests have been adapted and/or new tests have been added
- [x] Add a CHANGELOG.md entry
- [x] Update the README.md
- [x] Code style has been fixed via `composer fix-style`

Issue #683, opened 2020-10-22
Select * is preventing from groupping by for Interface types


EdwinDayot:
### Versions:
- graphql-laravel Version: 5.1.4
- Laravel Version: 8.0
- PHP Version: 7.4

### Description:

The generated SQL for a group by on an interface would raise an error as * is selected instead of the group by fields

### Steps To Reproduce:

```php
<?php

class MyQuery extends Query {

    public function type()
    {
        return Type::listOf(GraphQL::type('anInterface'));
    }

    public function resolve($root, $args, $context, $resolveInfo, $getSelectFields)
    {
        $fields = $getSelectFields();
        $select = $fields->getSelect();
        $with = $fields->getRelations();

        $myResults = MyModel::select($select)->with($with)->groupBy('my_field')->get();
    }
}
```

```php
<?php

class MyInterface extends InterfaceType {
    public function fields(): array
    {
        return [
            'my_field' => ['type' => Type::string()]
        ]
    }
}
```

```php
<?php

class MyType extends Type {
    public function fields(): array
    {
        return $interface->getFields();
    }
}
```

With this kind of configuration, the SQL request would result in:

```sql
select * from my_table group by my_field;
```

And of course, it would raise an SQL error: `Syntax error or access violation: 1055 Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column ...`

### Expectations

Only the required fields should be selected

illambo:
Maybe the problem is this https://github.com/rebing/graphql-laravel/blob/19ea3ed120b96c8c9b45398d0fb1ab1514860a4d/src/Support/SelectFields.php#L256

EdwinDayot:
I think it's just a part of a larger problem. Each Interface or Union should assume that every field defined in it are available in the database, but should use the alias defined in the type implementing the Interface or Union.

I have been working on it with the method 

https://github.com/rebing/graphql-laravel/blob/19ea3ed120b96c8c9b45398d0fb1ab1514860a4d/src/Support/SelectFields.php#L449

But it does not support aliases. (apparently)

illambo:
Great! May the fourth be with you :)

mfn:
Added the `SelectFields` label, seems like another shortcoming with this part.

crissi:
I am taking a look at this, it seems that relationships are not working well either...

crissi:
[...]

PR #1132, opened 2024-03-05
Relax PaginationType getPaginationFields typehint

jasonvarga:
## Summary

When upgrading a project from 8.0 to 9.0, I began to see errors around pagination:

```
TypeError: Rebing\GraphQL\Support\PaginationType::getPaginationFields(): Argument #1 ($underlyingType) must be of type GraphQL\Type\Definition\ObjectType, GraphQL\Type\Definition\InterfaceType given, called in /path/to/rebing/graphql-laravel/src/Support/PaginationType.php on line 22
```

We have queries where the `type` is a paginator with interfaces:

```php
class EntriesQuery extends Query
{
    public function type(): Type
    {
        return GraphQL::paginate(GraphQL::type('EntryInterface'));
    }
}
```

This has worked fine until v9 (specifically #1033) since the typehint of `PaginationType::getPaginationFields()` was changed to `ObjectType`.

This PR changes the typehint from an `ObjectType` definition to the parent `Type` definition, which covers both `ObjectType` and `InterfaceType`.

I'm not sure if there's a specific reason `ObjectType` was used, but this fixes our issue and all of this package's tests continue to pass.

---

Type of change:

Checklist:
- [ ] Update the README.md

### DIFF
--- a/src/Support/PaginationType.php
+++ b/src/Support/PaginationType.php
@@ -29,7 +29,7 @@ class PaginationType extends ObjectType
         parent::__construct($config);
     }
 
-    protected function getPaginationFields(ObjectType $underlyingType): array
+    protected function getPaginationFields(GraphQLType $underlyingType): array
     {
         return [
             'data' => [


jasonvarga:
> Also relaxing should not be a breaking change, so we would not need to do a major version bump, WDYT?

Nah I don't think it needs to be breaking.

> Can you please add a dedicated test, i.e. one which covers what you needed it for (support for interface type)?

Can do 👍  Thanks!

jasonvarga:
I've added tests to where the existing pagination test was.

There's an assertion on the generated sql that doesn't currently pass because of an unrelated issue, so I've left it in there but commented out: [`392c17d` (#1132)](https://github.com/rebing/graphql-laravel/pull/1132/commits/392c17d4010a6b8754a04f8fb9796887cb7b14e1#diff-6712f9e617d19cadd65cd4ee028839ff18daf5cc66d3baf527eb1e4246eb6c24R138-R143)

The sql _should_ be `select "posts"."id"` but is currently `select *` - I believe this is the same or similar issue as #683.

I think it's out of scope of this PR to fix it.

Anyway, the tests in this class would fail without the typehint change in `getPaginationFields`.

jasonvarga:
Thank you!

PR #1135, opened 2024-03-17
Add `authenticate` function to Field

edgarsn:
## Summary
This feature adds `authenticate()` function to `Field` used by Query and Mutation. This is very similar to existing `authorization()` function.

The alternative use case before this PR would be like this:

```php
<?php
namespace App\GraphQL\Queries;

use Auth;
use Closure;
use GraphQL\Type\Definition\ResolveInfo;
use Rebing\GraphQL\Error\AuthorizationError;

class UsersQuery extends Query
{
    public function authorize($root, array $args, $ctx, ResolveInfo $resolveInfo = null, Closure $getSelectFields = null): bool
    {
        if (Auth::guest()) {
            throw new AuthorizationError('Unauthenticated');
        }
    
        return Auth::user()->can_see_users;
    }

    // ...
}
```

The main problem is that this way authorize method is not really reusable, e.g. when Query uses Trait which has a implementation of `authorize()` to check whether user is authenticated. After this PR, separating `authenticate()` and `authorize()` that can be solved putting `authentication()` logic in Trait and leaving `authorize()` to check specific ability.

At the end, developers code may look like this:

**app/Traits/RequiresGraphQLAuthentication.php**
```php
<?php

namespace App\Traits;

trait RequiresGraphQLAuthentication 
{

    public function authenticate($root, array $args, $ctx, ResolveInfo $resolveInfo = null, Closure $getSelectFields = null): bool
    {
        // true, if logged in
        return ! Auth::guest();
    }
}
```

**app/GraphQL/Queries/UsersQuery.php**

```php
<?php
namespace App\GraphQL\Queries;

use Auth;
use Closure;
use App\Traits\RequiresGraphQLAuthentication;
use GraphQL\Type\Definition\ResolveInfo;
use Rebing\GraphQL\Error\AuthorizationError;

class UsersQuery extends Query
{
    use RequiresGraphQLAuthentication;

    public function authorize($root, array $args, $ctx, ResolveInfo $resolveInfo = null, Closure $getSelectFields = null): bool
    {
        return Auth::user()->can_see_users;
    }

    // ...
}
```

At the same time using `authenticate()` in a failure will return "Unauthenticated" while `authorize()` will throw "Unauthorized" and frontend could more easily  determine the error context by default.

---

Type of change:

Checklist:
- [x] Update the README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,7 @@ config/graphql.php
       - [Defining middleware](#defining-middleware)
       - [Registering middleware](#registering-middleware)
       - [Terminable middleware](#terminable-middleware)
+    - [Authentication](#authenticationAu)
     - [Authorization](#authorization)
     - [Privacy](#privacy)
     - [Query variables](#query-variables)
@@ -1244,6 +1245,60 @@ The terminate method receives both the resolver arguments and the query result.
 
 Once you have defined a terminable middleware, you should add it to the list of
 middleware in your queries and mutations.
+
+### Authentication
+
+For Query or Mutation authentication, we can use `authenticate()` function.
+
+```php
+namespace App\GraphQL\Queries;
+
+use Auth;
+use Closure;
+use GraphQL\Type\Definition\ResolveInfo;
+
+class UserFeedQuery extends Query
+{
+    public function authenticate($root, array $args, $ctx, ResolveInfo $resolveInfo = null, Closure $getSelectFields = null): bool
+    {
+        // true, if logged in
+        return ! Auth::guest();
+    }
+
+    // ...
+}
+```
+
+This function is called before `authorization()` which can be used to check whether authenticated user is authorized (has ability/permission) to use this Query or Mutation.
+If not provided, it will default to true - passing authenticate function and continuing to `authorize()` check.
+
+If user is not authenticated, an "Unauthenticated" error message will be thrown.
+
+You can also provide a custom error message when the authentication fails:
+
+```php
+namespace App\GraphQL\Queries;
+
+use Auth;
+use Closure;
+use GraphQL\Type\Definition\ResolveInfo;
+
+class UserFeedQuery extends Query
+{
+    public function authenticate($root, array $args, $ctx, ResolveInfo $resolveInfo = null, Closure $getSelectFields = null): bool
+    {
+        // true, if logged in
+        return ! Auth::guest();
+    }
+
+    public function getAuthenticationMessage(): string
+    {
+        return 'You are not authenticated';
+    }
+
+    // ...
+}
+```
 
 ### Authorization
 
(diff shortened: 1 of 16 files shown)


illambo:
Hi, thanks for the PR but I do not understand why not just integrate a simple custom [resolver middleware](https://github.com/rebing/graphql-laravel?tab=readme-ov-file#resolver-middleware) ?

illambo:
It would probably be interesting to bring both the authorization and validation layers as middleware but at the moment I have no way of being able to try this.
I wouldn't change the current implementation at the moment, currently what I was recommending was the use of an additional middleware. 
For example, since I prefer to first perform the authentication and authorization check before validation (for security reasons), I do not value the default [authorize](https://github.com/rebing/graphql-laravel?tab=readme-ov-file#authorization) field method (I let it always return to true) and instead implements a custom middleware (I attach an example below). 
Similarly one could implement and split the auth check from the permission check.

```php
abstract class Query extends \Rebing\GraphQL\Support\Query
{
    /**
     * @var string[]
     */
    protected $middleware = [
        \App\GraphQL\Middleware\Authenticate::class,
	// \App\GraphQL\Middleware\CheckPermissions::class,
    ];
}
```

```php
namespace App\GraphQL\Middleware;

use App\GraphQL\Errors\AuthorizationError;
use Closure;
use GraphQL\Type\Definition\ResolveInfo;
use Rebing\GraphQL\Support\Middleware;

class Authenticate extends Middleware
{
    /**
     * {@inheritdoc}
     *
     * @throws AuthorizationError
     */
    public function handle($root, array $args, $context, ResolveInfo $info, Closure $next)
    {
        if (! $this->authenticate($root, $args, $context, $info)) {
            throw AuthorizationError::unauthorized();
        }

        return $next($root, $args, $context, $info);
    }

    protected function authenticate($root, array $args, $context, ResolveInfo $info): bool
    {
        if (method_exists($root, 'authenticate')) {
            return $root->authenticate($root, $args, $context, $info);
        }

        return true;
    }
}
```

```php
ExampleQuery extends Query
{
    public const NAME = ‘example’;

    protected $attributes = [
        'name' => self::NAME,
    ];

    public function type(): Type
    {
        return GraphQL::type(ExampleType::NAME);
    }

    /**
     * @throws AuthorizationError
     */
    public function authenticate($root, array $args, $context, ResolveInfo $resolveInfo): bool
    {
        // todo: check auth ..

        return true;
    }

    // […]
}
```

edgarsn:
I kind a missed the resolver middleware feature. That's definately solving this. Maybe it would be good to mention some auth usage example under resolver middleware in docs?

I personally wouldn't like removing authorize fn. The authorize may vary query by query and moving to middleware wouldn't be so easy as it is right now imo.

illambo:
If u check the example implementation above you can see that is it possible customise query per query. Is it clear for you?

PR #1165, opened 2025-02-13
Bump thecodingmachine/safe to v3

the-pulli:
## Summary

Bumps `thecodingmachine/safe` to v3, and all corresponding deps to their current version to avoid deprecation warnings in PHP 8.4.
Removed the usage of `\Safe\sprintf` in favor of pure `sprintf`, because `sprintf` doesn't return `false` on error as of PHP 8.0.

Tested locally with PHP 8.3 and PHP 8.4 without issues.

Type of change:

Checklist:
- [ ] Update the README.md

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,11 +26,13 @@ jobs:
       fail-fast: false
       matrix:
         php: [8.1, 8.2, 8.3, 8.4]
-        laravel: [^10.0, ^11.0]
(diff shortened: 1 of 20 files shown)


the-pulli:
Nice to hear! Yeah, I just saw that. Didn't check it locally. Just ran the tests and thought I'll be good to go 🙈

the-pulli:
I just noticed that 🙃

PR #1195, opened 2025-11-06
Fix exponential time complexity in AliasArguments with circular type references

artem-schander:
## Problem

`AliasArguments::getAliasesInFields()` has exponential time complexity when processing input types with circular references, causing severe performance degradation in mutations.

**Root Cause:** The method traverses ALL possible fields in the type schema, not just fields present in the request data. With circular type references (e.g., `PageInput` ↔ `SectionInput`) and a depth limit calculated from request data, this creates exponential traversal.

## Evidence

Real-world example:
- **Before fix**: Mutation taking 11+ seconds
- **After fix**: Same mutation taking ~150ms  
- **Improvement**: 98.7% reduction in execution time

Debug logging revealed:
- 6 million recursive calls before timeout
- Depth reaching 12-13 levels with circular paths like: `newsletter.page.reference.page.redirect.page.translationOrigin.translationOrigin.translations.*.redirect.page...`

## Solution

Only traverse fields that exist in the actual request data. This prevents exponential explosion while preserving full alias functionality.

### Changes

**Modified signature:**
```php
protected function getAliasesInFields(array $fields, ?array $requestData = null, string $prefix = ''): array
```

**Key addition:**
```php
// Skip fields not present in actual request data
if ($requestData !== null && !array_key_exists($name, $requestData)) {
    continue;
}
```

**Recursive calls now pass data:**
```php
$pathAndAlias = $pathAndAlias + $this->getAliasesInFields(
    $type->getFields(),
    $fieldData,  // Pass actual data instead of all possible fields
    $newPrefix
);
```

## Testing

- ✅ Tested with deeply nested input types (11 levels)
- ✅ Tested with circular type references (PageInput ↔ SectionInput)
- ✅ Verified alias functionality preserved (deprecated field name mapping)
- ✅ Verified legitimate nested types work (e.g., `Page->translations` array)
- ✅ 124 production mutations tested successfully

## Impact

This fix affects any application with:
- Circular type references in input types (common in CMS/e-commerce)
- Deeply nested input data
- Alias usage for backward compatibility

## Backward Compatibility

✅ Fully backward compatible - the fix only changes the traversal strategy, not the API or functionality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ CHANGELOG
 
 [Next release](https://github.com/rebing/graphql-laravel/compare/9.13.0...master)
 
+## Fixed
+- Fix exponential time complexity in AliasArguments with circular type references [\#1195 / artem-schander](https://github.com/rebing/graphql-laravel/pull/1195)
+
 2025-11-30, 9.13.0
 ------------------
 


mfn:
@artem-schander can you rebase against master and fix the codstyle/phpstan stuff?

artem-schander:
there are many pre-existing code style issues

